PR #7, opened 2022-11-11
Feature/checksums

adairrr:
This change makes it so that the user can call `upload_if_needed()` on the contract interface and avoid a duplicate upload to the chain. This makes the experience for the user much better when an error is encountered during deployment and some part of it needs to restart.

It currently does NOT retrieve the checksum from the state logs, but rather by using `sha256` on the file itself to calculate the hash. This may or may not be desired, but avoids a query (if we could query that).


adairrr:
We may want to review how we're storing the information in the json file. As it stands, we have separate structures (`addresses`, `code_ids`, and now `checksums`) using the contract IDs as keys as opposed to a single contract ID key with those three values...

CyberHoward:
This was already somewhat implemented in the [daemon](https://github.com/Abstract-OS/BOOT/blob/main/boot-core/src/environments/daemon.rs#L62). Maybe you can see if you can scrape something from it. If not feel free to remove it. 

I don't think we should store the checksum information locally as we can easily retrieve it from a node. 
The address store is obviously required and the code-id store is required to store the latest code-ids for the specific contracts. 

Importantly there are times when you have uploaded a new code-id but haven't migrated the contract yet. If we were to store all the information under that contract key we must be very wary that we don't interpret it as "this contract has this code-id" or "this contract has this hash". 

This also relates to adding multiple deployments. ( #1 )
IMO by storing the state as below we keep all the functionality we want without running the risk of mis-interpreting any data. The blockchain is the most reliable data-source we have so let's make use of it. :p

```
network
└── chain-id
    ├── code-id
    └── addresses
        ├── deployment1
        │   ├── manager: "xxx"
        │   ├── proxy: "xxx"
        │   └── ...
        ├── deployment2
        │   └── ...
        └── ...
```

The `code-id` references the latest upload of the contract on the `chain-id` chain. If you want to detect if a new upload is required you can match that `code-id`'s hash (query) with your local one. 

If you then want to do automatic migrations you can get the code-ids for the contracts of your deployment (query) and match them with the code-ids in the state. If not equal, migrate. 

Local hashes can be found in the artifacts folder of the build optimizer (as you can see from the mentioned daemon impl) so definitely check that your "manual" hash matches the optimizer hash.

CyberHoward:
@adairrr is this something you want to work on or should I take it from here?

adairrr:
> @adairrr is this something you want to work on or should I take it from here?

@CyberHoward sure I've got it!

CyberHoward:
solves #1

PR #17, opened 2022-11-18
Update to trait-based interface

CyberHoward:
## :loudspeaker: Type of change

:white_small_square: Bugfix
:white_small_square: New feature
:white_small_square: Enhancement
:white_small_square: Refactoring
:white_small_square: Maintenance

## :pencil: Checklist

:white_small_square: Reviewed submitted code
:white_small_square: Added tests to verify changes
:white_small_square: Updated docs
:white_small_square: Verified on testnet


boot-core/examples/cw20.rs
```diff
@@ -1,26 +1,25 @@
-use boot_core::networks::juno::LOCAL_JUNO;
+use boot_core::networks::LOCAL_JUNO;
 
-use boot_core::instantiate_daemon_env;
+use boot_core::prelude::*;
```
adairrr, line 3: This is minor but would we be able to we avoid using glob imports? It makes the code harder to read because you don't necessarily know what's coming from where.

boot-cw-plus/src/cw1.rs
```diff
@@ -1,15 +1,15 @@
...
 
-use boot_core::{Contract, IndexResponse, TxHandler, TxResponse};
 use cosmwasm_std::Empty;
-use cw1_whitelist::msg::*;
+use cw1_whitelist::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
 use cw_multi_test::ContractWrapper;
-pub type Cw1<Chain> = CwPlusContract<Chain, ExecuteMsg, InstantiateMsg, QueryMsg, Empty>;
+
+#[boot_contract(InstantiateMsg, ExecuteMsg, QueryMsg, Empty)]
+pub struct Cw1;
```
adairrr, line 9: Woah this is awesome!

packages/boot-contract-derive/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+            panic!("Expected a contract endpoint type.");
+        };
+            type_path
+        })
+        .collect();
+
+    let exec = types_in_order[0].clone();
+    let init = types_in_order[1].clone();
+    let query = types_in_order[2].clone();
+    let migrate = types_in_order[3].clone();
```
adairrr, line 26: Is this ordering "enforced" by the type usages on lines #51-57?

boot-core/examples/cw20.rs
CyberHoward: Yeah I agree. That's why I use a separate prelude module. The reason is that a user needs to have the BootExecute, BootInstantiate, ... traits in scope before they can access the function calls on the contracts like `contract.execute()`.

So `prelude` only imports those specific traits to just make them available.

packages/boot-contract-derive/src/lib.rs
CyberHoward: Not yet, we could add a check for the name of the type if we want to enforce that.

CyberHoward: But a user would quickly see that they gave the types in the wrong order as a `.instantiate()` call would ask for their `QueryMsg` or whatever

PR #21, opened 2022-11-23
Execute message derive macro

CyberHoward:
## :loudspeaker: Type of change

:white_small_square: Bugfix
:white_small_square: New feature
:white_small_square: Enhancement
:white_small_square: Refactoring
:white_small_square: Maintenance

## :scroll: Description and Motivation

A derive macro that can be added to an `ExecuteMsg` struct to generate functions that can be called on the associated boot object. 

Example: 

```rust
#[cw_serde]
#[derive(ExecuteFns)]
pub enum ExecuteMsg{
    /// Freeze will make a mutable contract immutable, must be called by an admin
    Freeze {},
    /// UpdateAdmins will change the admin set of the contract, must be called by an existing admin,
    /// and only works if the contract is mutable
    UpdateAdmins { admins: Vec<String> },
    #[payable]
    Deposit {}
}

#[boot_contract(Empty,ExecuteMsg,Empty,Empty)]
struct Cw1

impl<Chain: BootEnvironment + Clone> Cw1<Chain> {
    pub fn test_macro(&self) {
        self.freeze().unwrap();
        self.update_admins(vec![]).unwrap(); 
        self.deposit(&[Coin::new(13,"juno")]).unwrap();
    }
}
```

## :pencil: Checklist

- Add flag for coins instead of optionals.

:white_small_square: Reviewed submitted code
:white_small_square: Added tests to verify changes
:white_small_square: Updated docs
:white_small_square: Verified on testnet

### DIFF
--- a/boot-core/src/lib.rs
+++ b/boot-core/src/lib.rs
@@ -13,6 +13,8 @@ pub mod state;
 mod tx_handler;
 
 // pub mod traits;
+pub use boot_contract_derive::boot_contract;
+pub use boot_fns_derive::ExecuteFns;
 pub use contract::Contract;
 pub use daemon::core::Daemon;
 pub use error::BootError;
@@ -20,8 +22,6 @@ pub use helpers::get_env_vars;
 pub use index_response::IndexResponse;
 pub use mock::core::Mock;
 pub use tx_handler::{TxHandler, TxResponse};
-pub use boot_contract_derive::boot_contract;
-pub use boot_exec_derive::ExecuteFns;
 /// Signals a supported execution environment
 pub trait BootEnvironment: TxHandler + Clone {}
 
--- a/boot-core/src/lib.rs
+++ b/boot-core/src/lib.rs
@@ -20,7 +20,8 @@ pub use helpers::get_env_vars;
 pub use index_response::IndexResponse;
 pub use mock::core::Mock;
 pub use tx_handler::{TxHandler, TxResponse};
-
+pub use boot_contract_derive::boot_contract;
+pub use boot_exec_derive::ExecuteFns;
 /// Signals a supported execution environment
 pub trait BootEnvironment: TxHandler + Clone {}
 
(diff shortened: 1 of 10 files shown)


boot-core/src/lib.rs
```diff
@@ -45,39 +46,4 @@ pub(crate) mod cosmos_modules {
     pub use cosmrs::proto::cosmos::tx::v1beta1 as tx;
     pub use cosmrs::proto::cosmwasm::wasm::v1 as cosmwasm;
     pub use cosmrs::proto::tendermint::abci as tendermint_abci;
-}
-
-// mod macro_dev {
-//     use terra_rust_script_derive::contract;
-
-//     #[derive(Clone, Debug, contract)]
```
adairrr, line 53: Is this what the `terra_rust_script_derive::contract` had done before?

packages/boot-exec-derive/src/lib.rs
```diff
@@ -0,0 +1,96 @@
+#![recursion_limit = "128"]
+
+extern crate proc_macro;
+use convert_case::{Case, Casing};
+use proc_macro::TokenStream;
+use quote::{format_ident, quote};
+use syn::{parse_macro_input, DeriveInput, Fields, Ident};
+
+#[proc_macro_derive(ExecuteFns,attributes(payable))]
+pub fn boot_execute(input: TokenStream) -> TokenStream {
```
adairrr, line 10: Will this be able to handle more advanced top-level structs like the `Middleware` ones with generics?

boot-core/src/lib.rs
CyberHoward: Yes and no, the old macro just returned the ExecuteMsg variant. This generates a trait and implements it for any contract with that ExecuteMsg. The `payable` attribute lets you specify which functions should accept a payment. All the generated functions can then be called on any boot-object (with any boot-compatible backend environment)

CyberHoward: I've been waiting with writing our testing infrastructure for this haha

packages/boot-exec-derive/src/lib.rs
CyberHoward: Nope this is just for regular contracts. We could augment it to fit our needs in the Abstract crates.

boot-core/src/lib.rs
adairrr: > Yes and no, the old macro just returned the ExecuteMsg variant. This generates a trait and implements it for any contract with that ExecuteMsg. The `payable` attribute lets you specify which functions should accept a payment. All the generated functions can then be called on any boot-object (with any boot-compatible backend environment) 

Yeah the payable attribute seems really versatile and smart contract authors can use it as a reference to add others ~

PR #66, opened 2023-03-31
Update `boot-cw-plus`

CyberHoward:
Rehaul the boot-cw-plus crate. 

It now uses the [abstract fork of cw-plus](https://github.com/Abstract-OS/cw-plus) which contains the derive macro's for generating the interface functions. 

A top-level `CwPlus` object exposes the `store_on` function to easily upload CW plus to any mock or daemon. 
An included executable is used to fetch the `.wasm` files for each of the contracts from the cw-plus repo. 

Small adjustments were made to the function derive marco to support `ExecuteMsg`s with an "execute" variant. 

We now also re-export the ibc-registry types required to easily query chain information when the `daemon` feature is enabled.


boot-cw-plus/cw-artifacts/cw1_subkeys.wasm
adairrr, line 1: Perhaps we should include these in a folder with the version of the cw-plus artifacts.

boot-cw-plus/src/cw_plus.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub cw20_base: Cw20Base<Chain>,
+    pub cw20_ics20: Cw20Ics20<Chain>,
+}
+
+impl<Chain: CwEnv> Deploy<Chain> for CwPlus<Chain> {
+    type Error = BootError;
+    type DeployData = Empty;
+
+    /// Deploy all cw-plus contracts to the given chain.
+    fn deploy_on(chain: Chain, _data: Empty) -> Result<Self, BootError> {
```
adairrr, line 21: Perhaps `upload_to` would be more appropriate. `deploy` implies that it is instantiated as well.

boot-cw-plus/src/contracts/cw1_subkeys.rs
```diff
@@ -0,0 +1,27 @@
+use boot_core::{contract, Contract, CwEnv};
+use cosmwasm_std::Empty;
+pub use cw1_subkeys::msg::{
+    ExecuteMsgFns as Cw1SubkeysExecuteMsgFns, QueryMsgFns as Cw1SubkeysQueryMsgFns,
+};
+use cw1_whitelist::msg::{ExecuteMsg, InstantiateMsg, QueryMsg};
```
adairrr, line 6: This should probably be `cw1_subkeys::msg`

boot-cw-plus/examples/cw-plus-mock.rs
```diff
@@ -0,0 +1,36 @@
+use boot_core::*;
+use boot_cw_plus::{Cw20QueryMsgFns, CwPlus};
+use cosmwasm_std::Addr;
+use cw20::Cw20Coin;
+
+// shows how to deploy CwPlus to a mock environment
```
adairrr, line 6: nit: these main functions should probably use `///` doc comment

boot-cw-plus/README.md
```diff
@@ -1,10 +1,28 @@
...
 
-Currently supports
-- CW-20
+## Usage
+
+add this to your `Cargo.toml`:
+
+```toml
+[dependencies]
+boot-cw-plus = { git = "" }
```
adairrr, line 11: `boot-cw-plus = { git = "https://github.com/Abstract-OS/BOOT" }`?

boot-cw-plus/src/contracts/cw1_subkeys.rs
CyberHoward: nice spot

boot-cw-plus/examples/cw-plus-mock.rs
CyberHoward: examples can't be imported so not something that's done 
i don't even think they are included when you run `cargo docs`

boot-cw-plus/src/cw_plus.rs
CyberHoward: In this specific case yes, not arguing that. I'm saying that we'd need to change the trait to reflect that in for it to be true in this specific case while for all other users `deploy_on` makes more sense (look at the WynDex integration and our own Abstract implementation of this trait)

CyberHoward:
If you want I can add a function to the trait `upload_to`

boot-cw-plus/src/cw_plus.rs
CyberHoward: will add an `upload_to` fn to the trait

boot-cw-plus/cw-artifacts/cw1_subkeys.wasm
CyberHoward: Yeah if someone requests it we can, people shouldn't use these WASMs in production.

PR #67, opened 2023-03-31
implement `CallAs` for Mock and Daemon environments.

CyberHoward:
as the title says, this allows us to use `.call_as()` in generic contexts!

### DIFF
--- a/boot-core/src/daemon/core.rs
+++ b/boot-core/src/daemon/core.rs
@@ -256,9 +256,8 @@ impl TxHandler for Daemon {
 impl<T: BootExecute<Daemon> + ContractInstance<Daemon> + Clone> CallAs<Daemon> for T {
     type Sender = Wallet;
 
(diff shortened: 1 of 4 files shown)


boot-core/src/daemon/core.rs
```diff
@@ -250,6 +253,21 @@ impl TxHandler for Daemon {
     }
 }
 
+impl<T: BootExecute<Daemon> + ContractInstance<Daemon> + Clone> CallAs<Daemon> for T {
+    type Sender = Wallet;
+
+    fn set_sender(&mut self, sender: &Self::Sender) -> &mut Self {
```
adairrr, line 259: Would prefer `with_sender` implying that it allows for chaining but lgtm!

adairrr (approved):

boot-core/src/daemon/core.rs
CyberHoward: can update it to remove the `&mut Self` return, will have a look in our codebase if we actually use the return ever.

PR #75, opened 2023-04-06
feature/tests progress

BHIIKTOR:
* added: speculoos to Cargo.toml
* added: boot-core/src/tests/unit_tests
* added: boot-core/src/tests/unit_tests/mock
* added: boot-core/src/tests/unit_tests/mock_state
* added: boot-core/src/tests/unit_tests/daemon_state
* modified + comments: check_file_validity in boot-core/src/daemon/state
* added: boot-core/gen_cov_test that generates coverage reports using grcov
* modified: .gitignore to include **/*.profraw just in case (from grcov)


boot-core/gen_cov_test
CyberHoward, line 1: I don't think this script is required (unless you want to generate the coverage manually). 

Here is the coverage report page: https://app.codecov.io/gh/abstract-os/BOOT
And here's the action that generates the reports: https://github.com/Abstract-OS/BOOT/blob/a42ce7d810eb0da1477aa75fe9daa3d58a80aa9d/.github/workflows/test.yml#L29

Lmk if you use this doesn't satisfy your requirements. Happy to hear how you use this script

boot-core/src/daemon/state.rs
```diff
@@ -147,29 +175,61 @@ impl DaemonState {
...
         };
+
+        // check validity
         state.check_file_validity();
+
+        // finish
         Ok(state)
     }
 
+    // maybe we shold rename this?
```
CyberHoward, line 186: sure, suggestions?

boot-core/src/daemon/state.rs
```diff
@@ -147,29 +175,61 @@ impl DaemonState {
...
     pub fn check_file_validity(&self) {
-        let file = File::open(&self.json_file_path).unwrap_or_else(|_| {
-            let file = File::create(&self.json_file_path).unwrap();
-            serde_json::to_writer_pretty(&file, &json!({})).unwrap();
-            File::open(&self.json_file_path).unwrap()
-        });
-        log::info!("Opening daemon state at {}", self.json_file_path);
```
CyberHoward, line 160: Would be nice to still have this logged!

boot-core/src/tests/unit_tests/daemon_state.rs
```diff
@@ -0,0 +1,49 @@
...
+        let options = DaemonOptionsBuilder::default()
+            .network(network)
+            .deployment_id(deployment_id)
+            .build()
+            .unwrap();
+
+        let daemon_state = DaemonState::new(options);
+
+        // start the daemon
+        let res = rt.block_on(daemon_state).unwrap();
```
CyberHoward, line 35: Since this line reaches out the the PISCO node I don't think it should be in unit test. 

Imo we should set up an integration test that spins up its own daemon is tests the daemon code with that because testnets + endpoints are very unreliable. 

Happy to hear your opinion and maybe we can jump on a call to discuss how we can implement that?

boot-core/src/tests/unit_tests/mock.rs
CyberHoward, line 1: Most of the library can and should be tested with the mock imo so great job on this!

boot-core/src/tests/unit_tests/mock.rs
```diff
@@ -0,0 +1,159 @@
...
+            .is_equal_to(&String::from("mint"));
+
+        let query_res = chain.query::<
+                cw20_base::msg::QueryMsg,
+                Response
+            >(&cw20_base::msg::QueryMsg::Balance {
+                address: recipient.to_string()
+            }, &contract_address).unwrap();
+
+        asserting("that query passed on correctly")
```
CyberHoward, line 151: if there is a `asserting!(..)` variant of this (notice the `!`) than prefer using that because it will add the line of the failed assertion to the test's output so you don't have to dig into your tests to find where it failed!

boot-core/src/tests/unit_tests/mod.rs
CyberHoward, line 1: For unit-tests in Rust the tests are normally placed in the file of the test's implementation. I.e. the mock state test can be placed in `boot-core/src/mock/state.rs`

That way you can very granularly test the code that is specifically in that file. Integration tests are put in their own folder. 

Here's an example of our Manager contract in Abstract. 
Unit tests: https://github.com/Abstract-OS/contracts/blob/47a5b00e20f2c535ff037958634e5eb8b04131e3/contracts/account/manager/src/commands.rs#L608
Integration tests: https://github.com/Abstract-OS/contracts/tree/main/contracts/account/manager/tests

boot-core/gen_cov_test
BHIIKTOR: the script allows me to check my progress on the spot and more quickly, without having to depend on doing a PR.
i can just skip having it in the repo

boot-core/src/daemon/state.rs
BHIIKTOR: uops, adding that back

boot-core/src/tests/unit_tests/mock.rs
BHIIKTOR: ok!, changing that. first time using speculoos

boot-core/src/tests/unit_tests/mod.rs
BHIIKTOR: some projects do it in the same file but also others split them up as much as posible.
i guess it all depends on coding style of the project.
given this comment ill assume that in abstract we prefer to do it as you say and move the code

boot-core/gen_cov_test
CyberHoward: Or just add it to `.gitignore` :)

boot-core/src/tests/unit_tests/mod.rs
CyberHoward: Yup, i'd prefer that! thanks!

boot-core/src/tests/unit_tests/daemon_state.rs
BHIIKTOR: totally agree that we should trash this and move the test to an integration test.
after i wrote the unit test i was left thinking that it was pointless but left it for review
so i could get feedback and see how to proceed in this type of occasions

boot-core/src/daemon/state.rs
BHIIKTOR: what about write_state_json?

boot-core/src/daemon/state.rs
CyberHoward: I don't see it in here yet?

boot-core/src/daemon/tx_resp.rs
```diff
@@ -140,43 +194,40 @@ impl From<StringEvent> for TxResultBlockEvent {
...
+    pub fn get_attributes(&self, key: &str) -> Vec<TxResultBlockAttribute> {
         self.attributes
             .iter()
             .filter(|attr| attr.key == key)
             .cloned()
             .collect()
     }
+
     /// return the first value of the first attribute that has the key 'key'
     pub fn get_first_value(&self, key: &str) -> Option<String> {
```
CyberHoward, line 206: We should rename this to `get_first_attribute_value`

boot-core/src/daemon/tx_resp.rs
```diff
@@ -104,20 +105,73 @@ impl From<TxResponse> for CosmTxResponse {
...
+        )))
+    }
 }
 
-impl From<Attribute> for TxResultBlockAttribute {
-    fn from(a: Attribute) -> Self {
+#[derive(Deserialize, Clone, Serialize, Debug)]
+pub struct TxResultBlockMsg {
+    pub msg_index: Option<usize>,
+    // pub log: Option<String>,
```
CyberHoward, line 158: Can remove this

CyberHoward (commented):
Looking good sir! Added a few minor comments. 

I think it makes sense to now test the `IndexResponse` trait. 

The way i'd go about it is by testing the trait for the `AppResponse`, then isolating the test from the `AppResponse` instantiation by creating test functions that take a trait object `&dyn IndexResponse`. 

Then what you can do  in the test is 

```rust
fn test_uploaded_code_id(&dyn IndexResponse) -> anyhow::Result<()> {
    // the actual assertions go here
}

#[test]
fn app_uploaded_code_id() {
   // create app response
   let app_resp = ...
   // call test function
   test_uploaded_code_id(&app_resp).unwrap()
}
```

And now we can re-use this code for the daemon `TxResponse` 

```rust
#[cfg(feature = "daemon")
mod daemon {

#[test]
fn uploaded_code_id() {
   // create TxResponse
   let tx_resp = ...
   // call test function
   test_uploaded_code_id(&tx_resp)
   }
}
```

It's then 10x easier for us to add tests to new objects in the future!

boot-core/src/daemon/contract.rs
```diff
@@ -7,31 +7,45 @@ use std::{
...
             log::info!("{} is already uploaded", self.id);
             Ok(None)
         } else {
+            log::info!("{} is not uploaded, uploading...", self.id);
             Some(self.upload()).transpose()
         }
     }
 
-    /// Returns a bool whether the checksum of the wasm file matches the checksum of the previously uploaded code
+    // Returns a bool whether the checksum of the wasm file matches the checksum of the previously uploaded code
```
CyberHoward, line 25: Yeah, while this probably will not fail, any network request could still fail and if it does then letting that error propagate up helps the dev identify the issue (his wifi going down/whatever). 

Also in a horror scenario the `latest_uploaded_code_id` could be completely wrong due to a dev using the wrong state file for example, which could also generate an error (like that the code-id doesn't exist)

boot-core/src/daemon/contract.rs
```diff
@@ -89,28 +103,48 @@ where
...
-    /// Calculate the checksum of the wasm file to compare against previous uploads
+    // Calculate the checksum of the wasm file to compare against previous uploads
+    // NOTE: Should we change this? (and maybe move it out of here?, perhaps to an utils file or something)
+    // if this function at the bottom returns the checksum of the file
+    // when the checksums.txt file is not found what is the difference?
+    // what i mean is why is relevant here to check for the existence of a checksums.txt file
+    // parse it and find our contract in it
+    // if the default way is to do it in the function anyway
+    // for example: what if the file is named *checksums_intermediate.txt*
+    // which is the default name rust-optimizer generates?
```
CyberHoward, line 114: checksums.txt is the default name for workspace optimizer.

CyberHoward: Hmmm i get your point. Could you verify that the hash of the digest is actually equal? if it is then sure we can remove that code! Hashing is a bit expensive but we don't care that much about performance.

boot-core/src/daemon/state.rs
```diff
@@ -9,10 +9,19 @@ use cosmwasm_std::Addr;
...
 use serde_json::{from_reader, json, Value};
-use std::{collections::HashMap, env, fs::File, rc::Rc, str::FromStr};
+use std::{collections::HashMap, env, fs::{File, OpenOptions}, rc::Rc, str::FromStr};
 use tonic::transport::{Channel, ClientTlsConfig};
 pub const DEFAULT_DEPLOYMENT: &str = "default";
 
+/*
+    the proper way of using DaemonOptions is using DaemonOptionsBuilder
+    here is an example of how:
+    let options = DaemonOptionsBuilder::default()
```
CyberHoward, line 19: You can use ``` ``` to put code into your doc comments to make them look nice! 
Give this a read when you have some time: https://doc.rust-lang.org/rust-by-example/meta/doc.html

boot-core/tests/contract.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    static mut DOCKER_CONTAINER_ID: Lazy<String> = Lazy::new(|| String::new());
+    static INIT: Lazy<Mutex<bool>> = Lazy::new(|| Mutex::new(false));
+
+    // const CW20_CONTRACT_WASM: &str = "/../boot-cw-plus/cw-artifacts/cw20_base.wasm";
+    const CW20_CONTRACT_WASM: &str = "/../boot-cw-plus/cw-artifacts/cw20_base.wasm";
+    const STATE_FILE: &str = "/tmp/boot_test.json";
+    const LOCAL_MNEMONIC: &str = "clip hire initial neck maid actor venue client foam budget lock catalog sweet steak waste crater broccoli pipe steak sister coyote moment obvious choose";
+
+    #[ctor]
```
CyberHoward, line 42: This code should remain in the `common/mod.rs` file as it can serve as the basis for all our integration tests.

boot-core/src/daemon/contract.rs
BHIIKTOR: 1) workspace optimizer and the rust optimizer have a different output filename?
https://github.com/CosmWasm/rust-optimizer/blob/main/optimize.sh
I had no clue

2) I already did some testing and asked GPT4 about it (for a quick answer) hehe

boot-core/src/daemon/contract.rs
```diff
@@ -89,28 +103,48 @@ where
         Ok(wasm_code_path)
     }
 
-    /// Calculate the checksum of the wasm file to compare against previous uploads
```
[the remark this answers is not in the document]
BHIIKTOR: Ok, ill use those, thanks for the head up!

in vs code It will show the first comment on top of whatever you are hovering (on mouse enter) if you press command (for mac), wont toggle if you already have the mouse over

but yea those are different from doc comments

boot-core/src/daemon/contract.rs
BHIIKTOR: Ok I see your point.

the function above `upload_if_needed` is not working as intended , because it's getting this error:

`upload_if_needed response: Err(
    DaemonError(
        Status(
            Status {
                code: Unknown,
                message: "not found",
                metadata: MetadataMap {
                    headers: {
                        "content-type": "application/grpc",
                        "x-cosmos-block-height": "1",
                    },
                },
                source: None,
            },
        ),
    ),
)`

im reversing the code to its original form and await more feedback on this

boot-core/src/daemon/state.rs
CyberHoward: *tripple backtick haha

boot-core/src/daemon/state.rs
```diff
@@ -9,19 +9,25 @@ use cosmwasm_std::Addr;
...
 /*
-    the proper way of using DaemonOptions is using DaemonOptionsBuilder
     here is an example of how:
     let options = DaemonOptionsBuilder::default()
         .network(LOCAL_JUNO)
         .deployment_id("v0.1.0")
         .build()
         .unwrap();
 */
+/// the proper way of using DaemonOptions is using DaemonOptionsBuilder
```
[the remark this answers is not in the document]
BHIIKTOR: that looks really cool!

of course im using rust-analyzer!

boot-core/tests/common/mod.rs
```diff
@@ -29,21 +63,24 @@ fn docker_container_setup() {
...
+    log::info!("Using CONTAINER_NAME: {}", container);
+    log::info!("Using STATE_FILE: {}", env::var("STATE_FILE").unwrap());
     log::info!(
-        "using LOCAL_MNEMONIC: {}",
+        "Using LOCAL_MNEMONIC: {}",
         env::var("LOCAL_MNEMONIC").unwrap()
     );
 
     if !*init {
+        ensure_container_removal();
```
CyberHoward, line 75: If you read the `CONTAINER_NAME` above you should pass it in here instead of re-reading it in the function.

boot-core/src/daemon/state.rs
```diff
@@ -126,7 +123,7 @@ impl DaemonState {
                 pub_address_prefix: network.bech32_prefix,
                 coin_type: network.slip44,
             },
-            id: network.chain_id.to_string(),
+            network_id: network.chain_id.to_string(),
```
CyberHoward, line 126: network_id = chain_id?

boot-core/src/daemon/networks/archway.rs
```diff
@@ -1,15 +1,15 @@
 use crate::networks::{ChainInfo, NetworkInfo, NetworkKind};
 
 pub const ARCHWAY_CHAIN: ChainInfo = ChainInfo {
```
CyberHoward, line 3: Can you rename all the `ChainInfo` to `NetworkInfo` and the other way around?

boot-core/src/daemon/networks/injective.rs
```diff
@@ -19,7 +19,7 @@ pub const INJECTIVE_1: NetworkInfo = NetworkInfo {
 
 /// https://docs.injective.network/develop/public-endpoints
 /// https://testnet.status.injective.network/
-pub const INJECTIVE_888: NetworkInfo = NetworkInfo {
+pub const INJECTIVE_888: ChainInfo = ChainInfo {
     kind: NetworkKind::Testnet,
```
CyberHoward, line 23: Can you rename this as well to `ChainKind` to keep things consistent?

boot-core/src/daemon/networks/injective.rs
```diff
@@ -1,12 +1,12 @@
-use crate::networks::{ChainInfo, NetworkInfo, NetworkKind};
+use crate::networks::{NetworkInfo, ChainInfo, NetworkKind};
 
-pub const INJECTIVE_CHAIN: ChainInfo = ChainInfo {
+pub const INJECTIVE_CHAIN: NetworkInfo = NetworkInfo {
```
CyberHoward, line 3: these `_CHAIN` can also become `_NETWORK`

PR #76, opened 2023-04-08
Integration testing for `Daemon` code

CyberHoward:
Adds an integration testing setup for testing the `Daemon` code of BOOT. 

These tests consist of the setup, located in `boot-core/tests/common/mod.rs` and the actual tests located in `boot-core/tests`. 

The setup code spins up a local juno daemon and sets the required environment variables for the integration test to proceed. 

The integration tests then connect to the daemon over gRPC and execute the tests. 
Currently these tests can only be ran in sequence as attempts to run them concurrently will result in account sequence errors.

# Test
Ensure your docker daemon is running an run `cargo test` to try it out!


CyberHoward:
used in #75

PR #77, opened 2023-04-11
Feature/new contract interface

Kayanski:
This PR aims ar providing higher level interfaces with simple definitions for user that start with BOOT and don't want to bother with a lot of the syntactic sugar needed to make the library work today

## Additions
This PR adds a `boot_contract` macro that allows for simpler `Contract Interface` Generation

## Ideas
We could even extend that to automatically derive the `ExecuteMsgFns`  and `QueryMsgFns` traits (but I don't think it's possible with rust as of today because the enum definitions are not included in the macro invocation)


Kayanski:
This macro is supposed to be indicated above each entry point function (instantiate, execute, query and migrate). It registers the endpoint as well as the message types. 

The instantiate function is treated a little differently because I think that this function is the one that defines the contract and thus it's where I define the underlying contract Struct

CyberHoward (commented):
Cool stuff! I don't quiet understand how this will be used by a user. Would it be possible to add a bit code examples. doesn't have to be compilable, just showcase how this works together. 

I'm guessing you have a mock set up to see your macro expansions? Feel free to include that in your branch as well while we work this out.

Kayanski:
> I'm also wondering if this change would make it impossible to create boot interfaces for contracts that are not under your control?
> 
> The current setup, while a bit verbose at times, does allow you to just import the messages of others and create an interface with them, allowing you to include those contracts into your setup without their explicit approval or a fork.

The way I see this change is an addition to how things are going right now. I'm not saying this change is the best solution and it solves all issues but you could use it as an additional way of instantiating the interface (that is less verbose on the user side)

CyberHoward:
Have you verified that ` std::env::var("CARGO_PKG_NAME").unwrap()` returns the name of the crate that's using the macro as a dependency instead of the macro crate itself? Never verified this myself. 

I like the idea! Think the only hard blocker will be figuring out if we want to feature flag these macros or not. 
Ideally we'd replace the `entrypoint` macro with our own as you eluded to but that might increase wasm sizes too much. Would be really interested in seeing a comparison. 

We have a fork of cw-plus with the `Fns` macros that you could easily use as a starting point: https://github.com/Abstract-OS/cw-plus
Remove the feature flag on the cw20 for example and point the boot source to your branch, include your macros, compile and compare to the released wasms that are included in the release on github.

Kayanski:
One issue I see for now (that needs to be fixed), is that we can't declare mutable DepsMut objects in the function (instantiate for instance), otherwise an error pop up ` patterns aren't allowed in function pointer types`. I don't know how to fix that error yet. A workaround exists to still use mutable variables, but that's not optimal.

CyberHoward:
can you do `let mut deps = deps` right after the function?

packages/boot-contract-derive/src/lib.rs
```diff
@@ -52,12 +59,149 @@ pub fn contract(attrs: TokenStream, input: TokenStream) -> TokenStream {
...
+
+        // We add the contract creation script
+        impl<Chain: ::boot_core::CwEnv> #name<Chain> {
+            pub fn new(contract_id: &str, chain: Chain) -> Self {
+
+                // We get the workspace dir
+                let workspace_dir = find_workspace_dir();
+
+                // We build the artififacts from the artifacts folder (by default) of the package
+                let file_path = &format!("{}/artifacts/{}.wasm", workspace_dir.display().to_string(), #wasm_name);
```
CyberHoward, line 154: when compiling with the arm optimizer the artifacts get a post-fix of `-aarch64.wasm`. Since a lot of devs use apple macbooks we should try to account for it.

packages/boot-contract-derive/src/lib.rs
```diff
@@ -52,12 +59,149 @@ pub fn contract(attrs: TokenStream, input: TokenStream) -> TokenStream {
...
+            impl<Chain: ::boot_core::CwEnv> ::boot_core::ContractInstance<Chain> for #name<Chain> {
+                fn as_instance(&self) -> &::boot_core::Contract<Chain> {
+            &self.0
+        }
+            fn as_instance_mut(&mut self) -> &mut ::boot_core::Contract<Chain> {
+                &mut self.0
+            }
+        }
+
+        fn find_workspace_dir() -> ::std::path::PathBuf{
```
CyberHoward, line 124: Wondering about this. If the code that uses this package is published and imported by someone, what will happen? 

Essentially you either need to ship the wasms with the package (current boot-cw-plus approach) or you need the consumer of the package to have the wasms for the package in their own artifacts dir.

There are also two different uses, either the dev wants to publish his contracts individually and add the wasm to it individually, or the dev has a coherent set of contracts and wants to ship a package with a wrapper (ex. `CwPlus`) and all the wasms for the contracts. Which one do you want to support? Is there a way to support both?

packages/boot-contract-derive/Cargo.toml
```diff
@@ -11,5 +11,11 @@ description = "Attribute macro for creating a contract interface."
 proc-macro = true
 
 [dependencies]
+convert_case = "0.6.0"
+cosmwasm-std.workspace = true
```
CyberHoward, line 15: there is actually no need to import these as a macro works on the AST level (before any of these crates are interpreted).

packages/boot-contract-derive/src/lib.rs
CyberHoward: Have a PR open for something similar: https://github.com/Abstract-OS/BOOT/pull/78

packages/boot-contract-derive/src/lib.rs
Kayanski: Here, we assume that the creator of the package ships the artifacts dir along with the crate. That is the approach you chose and I tried to respect that. I also think it is a sane approach for package creators, because the wasms are not heavy and you can check the wasm hash easily against an on-chain one if needed. 

To answer the second question, today this supports shipping a workspace (as this cw-plus version that is imported here) or a single contract repo. This allows for most simple implementations to create their interfaces easily and without too much boilerplate. 
For devs that want a wrapper with functions and handles they can develop it themselves (as for you boot-cw-plus version), because I see them as more experienced and able to ship such code. We could add in the future something to help wrap packages with nice function helpers

packages/boot-contract-derive/Cargo.toml
Kayanski: Done in https://github.com/Abstract-OS/BOOT/pull/77/commits/f4648aea95d6901898a9330a4b579ac25fb3be61

Kayanski:
To resolve the artifacts issue, I would introduce some complexity in the wasm checking process.
Today we are only checking for either : 
1.  A fixed file with a fixed filename (very strict°
2. A loose file name inside the artifacts contract

I think we should make solution 1 less strict and introduce a regex component to it. Instead of matching a precise path, we could also allow for some flexibility by introducing path patterns like `ARTIFACTS_DIR/*my_contract*.wasm/`

This would remove the `-arm64` issue as well, what do you think ?

CyberHoward:
That's a good suggestion. This path resolver code will be ran at runtime when the interface contract is constructed so it still doesn't fully solve the artifacts distribution issue I brought up.

Maybe we should support github release URLs as .wasm sources as well? That way we don't have to ship the artifacts in the packages.

packages/boot-contract-derive/src/lib.rs
CyberHoward: The workspace approach works as long as the dependency is a git url to the repo. If we were to package one of the contracts and publish it to crates.io then your setup would break because the worspace artifacts wouldn't be included in that crate.

CyberHoward: And that's flaky behavior that people won't expect

CyberHoward: Another alternative could be adding a build.rs file that copies the workspace wasm to a dir in the crate to ensure it's included in the crate but that's a bit overkill

CyberHoward:
I updated the way we handle paths for mocks in #88. It should give you more freedom into how you deal with the `wasm` paths. 

I also bumped into this [`Contract`](https://github.com/CosmWasm/cw-multi-test/blob/main/DESIGN.md#contracts) trait. I think your macro should be suited to implement the trait on the actual interfaces, as well as implementing `Uploadable<Mock>` and `Uploadable<Daemon>` of the [`Uploadable`](https://github.com/Abstract-OS/BOOT/blob/e28c83dde49646a630b1afd09976241a3d3fb07e/boot-core/src/interface_traits.rs#L100) trait.

PR #82, opened 2023-04-19
Update daemon queriers

CyberHoward:
This PR separates the different query functions into their own objects. This cleans up the interface for doing queries, both in our own code as well as in the scripting code. 

A user can now easily call as specific query using the daemon object. Example:  
```rust
daemon.query::<Node>().block_height().await?
```

Internally queries can be preformed as long as the dev has access to the `Channel` it wants to query over. 
```rust
Node::new(channel).block_height().await?
``` 

New queriers + methods can easily be added in the future.


BHIIKTOR:
I have quickly checked the changes, love them!
Going to add some tests for them
👍

BHIIKTOR (approved):
we are ready to go with this

PR #86, opened 2023-04-19
add wait time

tekmos:
added wait_time

### DIFF
--- a/boot-core/src/daemon/core.rs
+++ b/boot-core/src/daemon/core.rs
@@ -220,6 +220,13 @@ impl TxHandler for Daemon {
         Ok(())
     }
 
+    fn wait_time(&self, secs: u64) -> Result<(), DaemonError> {
+        self.runtime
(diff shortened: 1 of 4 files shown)


boot-core/src/daemon/core.rs
```diff
@@ -220,6 +220,13 @@ impl TxHandler for Daemon {
         Ok(())
     }
 
+    fn wait_time(&self, secs: u64) -> Result<(), DaemonError> {
```
adairrr, line 223: @CyberHoward think wait_seconds might be clearer to the dev? Ohterwise lgtm

tekmos: renamed to wait_seconds

CyberHoward:
I'm going to change the implementation because it can screw up people using it. 
1. Wait X time should also increment the block height in the mock environment, as that's what would happen in a real env.
2. Wait X time on the daemon should wait for a block with a timestamp that's past the current time + wait time because that's also how smart-contracts get their time data. (open to discussion)

PR #90, opened 2023-04-24
Added Generic type support for ExecuteFns and QueryFns

Kayanski:
This PR aims at introducing support for Generic Parameters for Message Enums.

Previously, such block were yielding a compilation error : 

```rust
#[cw_serde]
// Errors!!! (not recognizing the <> brackets)
#[derive(ExecuteFns)]
pub enum Cw1ExecuteMsg<T = Empty>
where
    T: Clone + fmt::Debug + PartialEq + JsonSchema,
{
    /// Execute requests the contract to re-dispatch all these messages with the
    /// contract's address as sender. Every implementation has it's own logic to
    /// determine in
    Execute { msgs: Vec<CosmosMsg<T>> },
}
...
// Errors !!!! (not recognizing the <> brackets)
#[contract(InstantiateMsg, ExecuteMsg<Empty>, QueryMsg<Empty>, Empty)]
pub struct Cw1Subkeys;
```

After this PR, such blocks will be permitted.

Keep in mind that in the second code snippet, the objects inside the `contract` macro should represent types and should not be generic.

## Changes

* Changed the `boot-fns-derive::execute_fns_derive` and `boot-fns-derive::query_fns_derive` functions to also include the explicit generics of the message Enums
* Changed the `boot_contract_derive::contract` macro to not parse simple args anymore but to parse comma punctuates list of Path

## Dependencies

This PR is linked to the following PR in the `cw-plus` repo : https://github.com/Abstract-OS/cw-plus/pull/2


boot-cw-plus/src/contracts/cw3_flex_multisig.rs
```diff
@@ -7,7 +7,7 @@ pub use cw3_flex_multisig::msg::{
     ExecuteMsgFns as Cw3FlexMultisigExecuteMsgFns, QueryMsgFns as Cw3FlexMultisigQueryMsgFns,
 };
 
-#[contract(InstantiateMsg, ExecuteMsg, QueryMsg, Empty)]
```
CyberHoward, line 10: I'm wondering, is there a way to keep the ExecuteMsg generic here? That would be impossible with our current traits right?

Kayanski: I don't really see how it could be compatible with  the current traits. We assume, as every contract, that the types used for the messages don't have generics. I don't see the use case for a generic parameter here ?

CyberHoward: Yeah you're right

Kayanski: Ok, so this seems possible to create such a contract structure for later instantiation : Something like this for a single generic parameter and could be adapted for more type parameters

```rust
pub struct GenericTestContract<Chain: ::boot_core::CwEnv, T>(::boot_core::Contract<Chain>, Option<std::marker::PhantomData<T>>);
```

CyberHoward: Hmm yeah, and the CwInterface 

```rust
impl <Chain, T> CwInterface for GenericTestContract<chain, T> {
    type ExecuteMsg = ExecuteMsg<T>
}
```
Does that work?

CyberHoward: Even if it would work I think we should keep it for another PR. 👍

Kayanski: Yes this works : 
image

CyberHoward: can you try adding a default to T?
`ExecuteMsg<T = Empty>`
That way we don't always have to provide the generic

Kayanski: Sure, this was only a macro expansion to show you what code it would need to generate to make that work

PR #103, opened 2023-04-26
Adding queries to all queriers

BHIIKTOR:
Adding all possible missing queries for:

* Node
* CosmWasm
* Bank
* Gov
* Staking
* Feegrant


cw-orc/src/daemon/queriers.rs
```diff
@@ -1,3 +1,188 @@
+//! # DaemonQuerier
+//!
+//! ## Usage examples
+//!
+//! Note: all following examples need the following code:
+//!
+//! ```rust
+//! use std::sync::Arc;
+//! use tokio::runtime::Runtime;
+//! let rt = Arc::new(Runtime::new().unwrap());
```
CyberHoward, line 10: This doesn't need to be `Arc` for these examples. 
You should tell the developer that he needs `Channel` and that he can get it by either constructing it or by calling an object that implements `ChannelAccess` like the Daemon.

cw-orc/src/daemon/queriers/bank.rs
```diff
@@ -47,4 +47,93 @@ impl Bank {
...
+            }
+        );
+        Ok(spendable_balances)
+    }
+
+    /// Query total supply in the bank
+    pub async fn total_supply(
+        &self,
+        pagination: Option<PageRequest>,
+    ) -> Result<cosmos_modules::bank::QueryTotalSupplyResponse, DaemonError> {
```
CyberHoward, line 73: Sometimes the Results that are returned from a query are simple wrappers around a value. In that case it's easier for the user of the querier that you already unwrap the value for them. 
Here for example you could return the total supply as just a number instead of returning the `QueryTotalSupplyResponse` type (which just wraps the number).

BHIIKTOR: ok, ill change the ones that are just wrapping basic data!

CyberHoward: Yeah I think that makes sense, altough it might be better to expose the pagination anyway. Can we add it as an optional parameter where it applies?

cw-orc/src/daemon/mod.rs
```diff
@@ -36,5 +36,6 @@ pub(crate) mod cosmos_modules {
     pub use cosmrs::proto::ibc::core::channel::v1 as ibc_channel;
     pub use cosmrs::proto::ibc::core::client::v1 as ibc_client;
     pub use cosmrs::proto::ibc::core::connection::v1 as ibc_connection;
+    pub use cosmrs::proto::ibc::applications::transfer::v1 as ibc_application;
```
CyberHoward: can this be ibc_transfer? 
ICA is also an ibc application

BHIIKTOR: yes, let me change it

PR #105, opened 2023-04-27
Added interface tests

Kayanski:
This PR aims at addiing more tests for proc_macros.

We might add creating contracts using generic types to this macro at some point


CyberHoward:
Everything seems to be failing and the tests don't really test much of the marco logic. Going to change it to draft.

CyberHoward:
Would like to have this done before we release so bumping

PR #107, opened 2023-05-02
Feature/examples

BHIIKTOR:
# Counter example

Still WIP, But im making this PR draft for review and comments in here.

Missing some more comments in the source code and a bit of refactoring(?), in the prod and dev functions at the bottom.
Not sure if the refactoring is good for the purpose of this example, where we might need to difference some
of the Mock and Daemon flows.

Now I need to add the quick start markdown to the book using this code as reference.
This markdown will be focus in other more peripheral stuff, as in requirements and environment setup


examples/counter/src/contract.rs
```diff
@@ -0,0 +1,294 @@
...
+    COUNT.save(deps.storage, &Count(msg.conf.initial_value))?;
+    Ok(Response::default().add_attribute("action", "migrate"))
+}
+
+// Now that we have setup for our contract entry points, We can continue to the next step.
+// This is where more of the magic of cw-orchestrator occurs
+// In this case we will prepare a trait for our two scenarios Mock and Daemon
+// Daemon is our production scenario, deploying to a real blockchain, be it a local testnet, a tesnet our a mainnet
+// and Mock is our development scenario, used for unit testing and fine tuning our contract with speed
+trait CounterWrapper<T: TxHandler> {
```
CyberHoward, line 115: Can you explain why you create a trait for this and why it has this structure? 
The `interface` macros create a struct that is essentially already a wrapper. 

How will this help devs to understand cw-orch?

BHIIKTOR: hmm I honestly dont think this will help devs understand cw-orch more.
Added it to keep the example DRY.
What would you rather have here?, just repeat code in the implementations below?

I also think that maybe this pattern is going to be a common approach for many devs.
Because they dont know that or even care?, and will try to keep their code DRY too or just follow a similar approach...
IDK

PR #108, opened 2023-05-03
Fix `Uploadable` implementation

CyberHoward:
Rework the `Uploadable` trait to be chain-generic. This is required to enable functions generic over `Chain: CwEnv` to be able to run `.upload()` on their contract. 

- move `.upload` from `TxHandler` into `ChainUpload` trait to allow for a customized implementation for `Mock<Empty,Empty>`. 
- add `ChainUpload` to CwEnv.


Kayanski (commented):
Ok.
So for implementing this for a custom contract (without the interface macro), we need to implement Uploadable generically AND then implement it specifically for Daemon ? That sounds a little heavy, but I don't see how we can do better, because we have to specify the functions and paths anyway

PR #110, opened 2023-05-08
Feature/add sudo reply migrate to interface

Kayanski:
This PR aims at allowing the integration of sudo, reply and migrate msgs to the `interface` macro

This creates default function getters for those optional functions and overrides them if necessary, when registering the endpoints via the usual #[interface].

## Left to do

- Simplify the `ContractWrapper` logic if possible. Today, the nested if structure defined in the `interface` macro is not very nice. However, I don't see how to do better with the constraints ContractWrapper introduces.
- Should the sudo endpoint be called directly ? I'm not sure how this endpoint works, I heard, only the Blockchain can call this endpoint so maybe not ?


packages/cw-orch-contract-derive/src/lib.rs
```diff
@@ -350,12 +363,77 @@ pub fn interface(_attrs: TokenStream, mut input: TokenStream) -> TokenStream {
         impl ::cw_orch::Uploadable<::cw_orch::Mock> for #name<::cw_orch::Mock>{
             fn source(&self) -> <::cw_orch::Mock as ::cw_orch::TxHandler>::ContractSource{
```
Kayanski, line 364: This implementation is not optimal, but can we do better ?

Kayanski (commented):
Here is a gist of a implementation test : https://gist.github.com/Kayanski/b1c9d63911d55581e8908437a53d359e

Kayanski:
> Here is a gist of a implementation test : https://gist.github.com/Kayanski/b1c9d63911d55581e8908437a53d359e

This expands to : https://gist.github.com/Kayanski/0969ad5d5619fa181400dc69ac6d846e

CyberHoward:
Yeah this does seem rather expressive, is it not possible to work with re-assignements to get rid of the nested structure and Box it after that? 

Another alternative that I've floated earlier is implementing [`Contract`](https://docs.rs/cw-multi-test/0.16.4/cw_multi_test/trait.Contract.html) for the struct itself, bypassing the ContractWrapper entirely as the upload expects a `Box<dyn Contract>` which is a trait-object.

Kayanski:
> Yeah this does seem rather expressive, is it not possible to work with re-assignements to get rid of the nested structure and Box it after that?

Not really because the generic types are different when including a ẁith_sudo`, `with_reply`, from the standard `ContractWrapper` ...
> 
> Another alternative that I've floated earlier is implementing [`Contract`](https://docs.rs/cw-multi-test/0.16.4/cw_multi_test/trait.Contract.html) for the struct itself, bypassing the ContractWrapper entirely as the upload expects a `Box<dyn Contract>` which is a trait-object.

This should be the way to go yes

PR #113, opened 2023-05-08
Fix queriers' documentation and improve their UX

BHIIKTOR:
# Queriers fixes and UX improvements

WIP, not ready for review yet.

## Ideas gathered from devs feedback from twitter

1) Available balance
2) Highest amount staked to a validator
3) Highest amount currently unbonding from a validator
4) Quick and gas effective way to get total staked for an account across all validators
5) Check pending rewards
6) Check authz grants for a given contract

I believe 1 is already there, while 6 probably cant be done, and 4 is very relative to what is gas effective 🤔 
need to check how to do 5 🤔


CyberHoward:
Whoa, the description if this PR completely changed, discussion like this should be additive. Also I don't think we need complex queries like the ideas on twitter, while they are cool for specific use-cases we want cw-orch to be a generic tool that devs can use to build out these use-cases themselves. 

Lets wrap this PR up, there's other stuff to do.

PR #120, opened 2023-05-12
Add asynchronous DaemonAsync

CyberHoward:
Split out the asynchronous daemon interface from the synchronous one. 

This will help us improve our ability to apply cw-orch to async use-cases as well as giving us the groundwork for a multi-threaded implementation. 

- [x] Create DaemonAsync
## For future work
- [ ] Create AsyncTxHandler for mapping the async interfaces
- [ ] Create AsyncCwOrcX traits for contract interfaces
- [ ] Add these trait implementations to the macros


CyberHoward:
@Kayanski does the current impl (daemon queriers and low-level execution) fit your requirements or do you need the other unchecked boxes to be implemented before it's of use to you?

Kayanski:
> @Kayanski does the current impl (daemon queriers and low-level execution) fit your requirements or do you need the other unchecked boxes to be implemented before it's of use to you?

I only need the daemon creation via the builder to be async for my purposes. The other boxes would make working interchain faster but that's already great !

CyberHoward:
okey, then i'll push that out and get this merged as it is.

cw-orch/src/daemon/core.rs
```diff
@@ -139,98 +126,78 @@ impl TxHandler for Daemon {
...
-            .block_on(self.query::<Node>().block_height())?;
+    pub async fn wait_blocks(&self, amount: u64) -> Result<(), DaemonError> {
+        let mut last_height = self.query_client::<Node>().block_height().await?;
         let end_height = last_height + amount;
 
         while last_height < end_height {
-            // wait
-            self.rt_handle
-                .block_on(tokio::time::sleep(Duration::from_secs(4)));
+            tokio::time::sleep(Duration::from_secs(4)).await;
```
CyberHoward, line 172: There should be a way for us to get the average block time. Maybe wait for 2 blocks, get the average time and bump it by 20%? 

Because block times can vary a lot.

cw-orch/src/daemon/core.rs
```diff
@@ -254,36 +222,24 @@ impl ChainUpload for Daemon {
...
 
         let code_id = result.uploaded_code_id().unwrap();
 
         // wait for the node to return the contract information for this upload
         let wasm = CosmWasm::new(self.channel());
-        while self.rt_handle.block_on(wasm.code(code_id)).is_err() {
-            self.rt_handle
-                .block_on(tokio::time::sleep(Duration::from_secs(6)));
+        while wasm.code(code_id).await.is_err() {
+            tokio::time::sleep(Duration::from_secs(6)).await;
```
CyberHoward, line 234: Here same with the upload, we want to wait approx 1 block each time.

CyberHoward: see #125

cw-orch/src/daemon/core.rs
CyberHoward: see #125

PR #123, opened 2023-05-15
Add pre-push hook

CyberHoward:
Add a pre-push hook that performs formatting and clippy checks.


CyberHoward:
@Kayanski you can install it with `./hooks/pre-push.sh --install`

PR #131, opened 2023-05-18
Added generics test for the `interface` macro

Kayanski:
This PR aims at adding generics tests for the `interface` macro.
I have created another contract for using their endpoint and added a generic `ExecuteMsg`


CyberHoward:
can't we just have one mock contract?

PR #132, opened 2023-05-18
Refactor `DaemonState` to include chain registry `ChainData`.

CyberHoward:
This PR cleans up the `DaemonState` struct by removing all the chain-specific data and keeping it in a `chain_data` field that holds a `ChainData` struct, imported from chain registry. This way we can remove the owned chain data structs and we improve readability and maintainability.

### DIFF
--- a/cw-orch/src/daemon/chain_info.rs
+++ b/cw-orch/src/daemon/chain_info.rs
@@ -1,0 +1,159 @@
+use std::env;
+
+use serde::{Deserialize, Serialize};
+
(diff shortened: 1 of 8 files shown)


CyberHoward:
@Kayanski this PR cleans up some of the chain data in the DaemonState, would you want this merged before or after interchain?

Kayanski:
> @Kayanski this PR cleans up some of the chain data in the DaemonState, would you want this merged before or after interchain?

Let's do before, this will clean things up there too :)

PR #145, opened 2023-06-13
Add Injective Network support

CyberHoward:
Adds support for the injective network


cw-orch/src/daemon/sync/core.rs
```diff
@@ -66,6 +66,11 @@ impl Daemon {
     pub fn channel(&self) -> Channel {
         self.state().grpc_channel.clone()
     }
+
```
Kayanski, line 69: Added this back because the user of daemon doesn't have access to the underlying sender

cw-orch/src/daemon/networks/injective.rs
```diff
@@ -23,8 +23,8 @@ pub const INJECTIVE_888: ChainInfo = ChainInfo {
     kind: ChainKind::Testnet,
     chain_id: "injective-888",
     gas_denom: "inj",
-    gas_price: 0.025,
-    grpc_urls: &["https://testnet.grpc.injective.network:443"],
+    gas_price: 500_000_000.0,
```
CyberHoward, line 26: lol is this correct?

cw-orch/src/daemon/sender.rs
```diff
@@ -125,7 +127,8 @@ impl Sender<All> {
         );
 
         let auth_info =
-            SignerInfo::single_direct(Some(self.private_key.public_key()), sequence).auth_info(fee);
+            SignerInfo::single_direct(Some(self.cosmos_private_key().public_key()), sequence)
```
CyberHoward, line 130: Wow so the public key is derived from the cosmos private key but the signing requires a different protocol? That's interesting.

cw-orch/src/daemon/tx_builder.rs
```diff
@@ -122,16 +123,19 @@ impl TxBuilder {
             sequence
         );
 
-        let auth_info = SignerInfo::single_direct(Some(wallet.private_key.public_key()), sequence)
-            .auth_info(fee);
+        let auth_info = SignerInfo {
+            public_key: wallet.private_key.get_signer_public_key(&wallet.secp),
+            mode_info: ModeInfo::single(SignMode::Direct),
```
CyberHoward, line 128: Have you checked out `SignMode::Eip191`? Could possibly do the signing that's currently implemented in `proto/injective.rs`

cw-orch/src/daemon/networks/injective.rs
Kayanski: Yes, inj has 18 decimals

cw-orch/src/daemon/sender.rs
Kayanski: Ah, no didn't notice we also have to change things here

cw-orch/src/daemon/networks/injective.rs
CyberHoward: oh wow, okey!

cw-orch/src/daemon/tx_builder.rs
Kayanski: Will try that

Kayanski: It doesn't seem to work either

CyberHoward: okey, np, just wanted to check.

PR #146, opened 2023-06-14
Refactor tx construction and add insufficient fee fallback

CyberHoward:
Refactor the tx construction to be seperated from the sender object. This allows for more flexibly tx execution, which is applied in the case when the fee of a tx is not sufficient. In this case the returned log is parsed and the suggested fee is used.


CyberHoward:
Would like to get your feedback on this @Kayanski 
It doesn't work yet 😔

PR #147, opened 2023-06-14
Feature/add mock querier

Kayanski:
This PR aims at adding a mock querier that queries on-chain data for unit-testing contracts.
This can't be state-changing, but this allows one to query the actual on-chain test

### DIFF
--- a/cw-orch/src/mock/mock_querier.rs
+++ b/cw-orch/src/mock/mock_querier.rs
@@ -1,13 +1,13 @@
+use crate::daemon::queriers::CosmWasm;
+use crate::daemon::queriers::DaemonQuerier;
+use crate::prelude::queriers::Bank;
 use cosmwasm_std::AllBalanceResponse;
 use cosmwasm_std::BalanceResponse;
-use crate::daemon::queriers::DaemonQuerier;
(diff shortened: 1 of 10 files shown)


cw-orch/src/mock/mock_querier.rs
```diff
@@ -0,0 +1,261 @@
...
+        storage: MockStorage::default(),
+        api: MockApi::default(),
+        querier: custom_querier,
+        custom_query_type: PhantomData,
+    }
+}
+
+/// Querier struct that fetches queries on-chain directly
+pub struct WasmMockQuerier {
+    base: MockQuerier<Empty>,
```
CyberHoward, line 57: is this base required anywhere? If not, is there a scenario where it might be? If yes, expose it with `pub` or a getter. If not you can remove it.

CyberHoward (changes requested):

cw-orch/src/mock/mock_querier.rs
Kayanski: https://github.com/AbstractSDK/cw-orchestrator/pull/147/commits/1398297fe7ba043e11ec874b01616c53f9536c38

PR #148, opened 2023-06-20
[Fix] Remove QueryMsg variant doc attributes before function construction.

CyberHoward:
Allows for doc-comments on the `QueryMsg` variant fields when using the `QueryFns` macro.


Buckram123:
❤️ EPIC! Fixes CronCats/abstract-croncat-app#13

PR #150, opened 2023-07-05
Update/add osmosis test tube

Kayanski:
This PR aims at providing Osmosis-test-tube as an additional chain for cw-orch. 
This new chain is very similar to cw-multi-test but tests against actual on-chain logic.
Basic counter tests have been provided


cw-orch/Cargo.toml
```diff
@@ -24,7 +24,7 @@ name = "daemon_test"
 required-features = ["daemon"]
 
 [features]
-default = []
```
CyberHoward, line 27: we shouldn't have it default

cw-orch/src/test_tube/core.rs
```diff
@@ -0,0 +1,336 @@
...
+    environment::TxHandler,
+    error::CwOrchError,
+    state::{ChainState, StateInterface},
+};
+
+use crate::mock::MockState;
+
+pub use osmosis_test_tube;
+
+/// Wrapper around a cw-multi-test [`App`](cw_multi_test::App) backend.
```
CyberHoward, line 40: These doc comments should be updated

cw-orch/src/test_tube/core.rs
```diff
@@ -0,0 +1,336 @@
...
+                denom: c.denom,
+            })
+            .collect();
+        Ok(amount)
+    }
+}
+
+impl TestTube<MockState> {
+    /// Create a mock environment with the default mock state.
+    pub fn new(init_coins: Vec<Coin>) -> Self {
```
CyberHoward, line 185: To who are these init coins minted? Add some doc comment on that

cw-orch/src/test_tube/response.rs
CyberHoward, line 1: This file seems empty

CyberHoward (changes requested):
Wow, that's so little code for an integration, pretty awesome. 

Is there any pre-setup required to executing with this? I.e. is the binary included in the package or does the user need to install anything before they can run this? If so it should be clearly documented somewhere.

CyberHoward:
Oh and can we rename TestTube to OsmosisTestTube? I think the idea of test-tube is to allow other chains to use it as well, so there could be a NeutronTestTube, etc.

README.md
```diff
@@ -205,6 +205,10 @@ impl<Chain: CwEnv> Example<Chain> {
 }
 ```
 
+### Testing with OsmosisTestTube
+
+OsmosisTestTube is available for testing in cw-orchestrator. In order to use it, you may need to install [clang] and [go] to compile the osmosis blockchain that serves as the backend for this env.
```
CyberHoward, line 210: Is the compilation done by the crate or is there other setup? A link to the test-tube repo would be great here as well.

CyberHoward:
And could you add a usage example in /cw-orch/examples ? You need to add it to the cargo.toml as well and set the feature requirement

Kayanski:
> 

Done

CyberHoward (approved):

PR #152, opened 2023-07-09
Separated packages of cw-orch

Kayanski:
This PR aims at separating the daemon logic from its integration to cw-orch.
This allows crates to depend on the daemon implementation without depending on the whole cw-orch crate


CyberHoward:
Should we move this to packages?

Buckram123:
This PR also separates environment and mock packages alongside with daemon
- [x] Environment
- [x] Daemon
- [x] Mock
- [ ] Fix docs

cw-orch-daemon/src/keys/private.rs
```diff
@@ -134,12 +127,7 @@ impl PrivateKey {
         secp: &Secp256k1<C>,
     ) -> Option<SignerPublicKey> {
         if self.coin_type == ETHEREUM_COIN_TYPE {
-            #[cfg(feature = "eth")]
             return Some(self.get_injective_public_key(secp));
-            panic!(
-                "Coin Type {} not supported without eth feature",
-                ETHEREUM_COIN_TYPE
-            );
         }
```
adairrr, line 131: Is there a reason we are maintaining this logic even without the eth feature?

cw-orch-mock/src/core.rs
```diff
@@ -298,46 +307,32 @@ impl<S: StateInterface> TxHandler for Mock<S> {
...
-        self.as_instance_mut().chain.sender = sender.clone();
-    }
-
-    fn call_as(&self, sender: &Self::Sender) -> Self {
-        let mut contract = self.clone();
-        contract.set_sender(sender);
-        contract
-    }
-}
-
```
adairrr, line 340: IS there a blanketed impl on this? Why was it removed?

cw-orch-daemon/src/state.rs
```diff
@@ -26,6 +26,9 @@ pub struct DaemonState {
     pub chain_data: ChainData,
 }
 
+#[derive(Clone, Debug)]
+pub struct RcDaemon(pub Rc<DaemonState>);
```
adairrr, line 30: Curious of the reasoning behind calling it `RcDaemon` and not `RcDaemonState`?

cw-orch-environment/src/contract/interface_traits.rs
```diff
@@ -181,13 +181,16 @@ impl<T: ContractInstance<Chain> + Uploadable, Chain: CwEnv> CwOrchUpload<Chain>
...
     /// Clones the contract interface with a different sender.
-    fn call_as(&self, sender: &Self::Sender) -> Self;
+    fn call_as(&self, sender: &<Chain as TxHandler>::Sender) -> Self {
+        let mut contract = self.clone();
+        contract.set_sender(sender);
+        contract
+    }
 }
+
+impl<T: CwOrchExecute<Chain> + ContractInstance<Chain> + Clone, Chain: CwEnv> CallAs<Chain> for T {}
```
adairrr, line 196: Nice.

cw-orch-daemon/README.md
```diff
@@ -0,0 +1 @@
+../README.md
```
adairrr, line 1: We should include some details as to what the package is, or at least link it dynamically with md `See [top-level README] (../README.md)` (with no space)

cw-orch-daemon/src/queriers/cosmwasm.rs
```diff
@@ -81,18 +81,14 @@ impl CosmWasm {
         &self,
         address: impl Into<String>,
         pagination: Option<PageRequest>,
-    ) -> Result<Vec<cosmos_modules::cosmwasm::Model>, DaemonError> {
+    ) -> Result<cosmos_modules::cosmwasm::QueryAllContractStateResponse, DaemonError> {
```
adairrr, line 84: If we're changing the response here so that the user is aware of the pagination, we should do it everywhere (most of the queriers that accept pagination "unwrap" the sub response).

cw-orch-mock/src/core.rs
```diff
@@ -489,7 +484,7 @@ mod test {
         let contract_id = "my_contract";
         let code_id = 1u64;
         let address = &Addr::unchecked(BALANCE_ADDR);
-        let mut mock_state = Rc::new(RefCell::new(MockState::new()));
+        let mut mock_state = RcState(Rc::new(RefCell::new(MockState::new())));
```
adairrr, line 487: We could have a nice `new` fn for `RcState` so that callers can do `RcState::new(MockState::new())`

cw-orch-environment/src/contract/paths.rs
```diff
@@ -153,12 +155,7 @@ mod artifacts_dir {
                         None
                     }
                 })
-                .ok_or_else(|| {
-                    CwOrchError::StdErr(format!(
-                        "Could not find wasm file with name {} in artifacts dir",
-                        name,
-                    ))
-                })?;
+                .ok_or_else(|| CwEnvError::WasmNotFound(name.to_owned()))?;
```
adairrr, line 158: It would be great if this error also included the path in which it was not found, as I've run into many a time where I wasn't sure if it was automatic or env or whatever.

cw-orch-daemon/src/state.rs
adairrr: And secondly, what's the point in wrapping it?

cw-orch-daemon/src/core.rs
```diff
@@ -220,37 +218,6 @@ impl DaemonAsync {
         })
     }
 
-    /// Upload a contract to the chain.
-    pub async fn upload(
```
adairrr, line 224: DaemonAsync can't implement `TxHandler` because the async fns... hence why these were implemented manually. Why are we removing `upload` ?

cw-orch-daemon/src/core.rs
```diff
@@ -51,13 +49,13 @@ use tonic::transport::Channel;
     ## Querying
```
adairrr, line 50: Nit: The comments above with `TxHandler` are incorrect for this struct.

adairrr (changes requested):
Nice! This should really make the package a lot more maintainable.

One request would be to have the networks in a common place so that if we implemented a wasm-daemon (one that either leveraged web-grpc or rpc) then we could use the networks without enabling `daemon`.

cw-orch-mock/src/core.rs
Buckram123: https://github.com/AbstractSDK/cw-orchestrator/pull/152#discussion_r1268616244
Yep!

cw-orch-daemon/src/core.rs
Buckram123: I guess DaemonAsync::upload was unused: https://github.com/AbstractSDK/cw-orchestrator/blob/36ff755de436ac8d37da52ebba550c1e62f0357f/cw-orch-daemon/src/sync/core.rs#L120-L123

cw-orch-daemon/src/core.rs
Buckram123: > Nit: The comments above with `TxHandler` are incorrect for this struct.

Aha, wrong path and it doesn't implement TxHandler. Updated it!

cw-orch-daemon/src/queriers/cosmwasm.rs
Buckram123: I guess this change was aiming to unify Cosmawm::* functions responses.

cw-orch-daemon/src/keys/private.rs
```diff
@@ -134,12 +127,7 @@ impl PrivateKey {
         secp: &Secp256k1<C>,
     ) -> Option<SignerPublicKey> {
         if self.coin_type == ETHEREUM_COIN_TYPE {
-            #[cfg(feature = "eth")]
             return Some(self.get_injective_public_key(secp));
-            panic!(
-                "Coin Type {} not supported without eth feature",
-                ETHEREUM_COIN_TYPE
-            );
```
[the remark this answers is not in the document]
Buckram123: Got it back

cw-orch-daemon/src/core.rs
Buckram123: Got it back, now it uses upload directly from DaemonAsync as well!

adairrr (approved):
Nice!

cw-orch-daemon/src/queriers/cosmwasm.rs
Kayanski: Yes. This is needed if you can't query the whole contract state at once (for instance if you have a lot of state). 
I agree that pagination should always be reachable in queries, this reassures developpers. @CyberHoward do you agree ?

Kayanski:
@Buckram123 We may need to have all the packages in the packages folder and leave cw-orch as the only crate in the root dir

Buckram123:
> @Buckram123 We may need to have all the packages in the packages folder and leave cw-orch as the only crate in the root dir

Do you think we should remove directories prefixes under "packages/cw-orch-*" ( without renaming package name)

Kayanski:
> > @Buckram123 We may need to have all the packages in the packages folder and leave cw-orch as the only crate in the root dir
> 
> Do you think we should remove directories prefixes under "packages/cw-orch-*" ( without renaming package name)?

I prefer it with the prefixes, in order not to mix things up in the names and imports. But I'm not sure

cw-orch-daemon/src/queriers/cosmwasm.rs
CyberHoward: I agree but from my memory the `Response` struct only contained the inner `Vec<...>` so there's no information loss in unwrapping it. If there is then we should return the `Response` itself for sure.

CyberHoward:
> > > @Buckram123 We may need to have all the packages in the packages folder and leave cw-orch as the only crate in the root dir
> > 
> > 
> > Do you think we should remove directories prefixes under "packages/cw-orch-*" ( without renaming package name)?
> 
> I prefer it with the prefixes, in order not to mix things up in the names and imports. But I'm not sure

Having the dir name = package name is best practice.

.github/workflows/check.yml
```diff
@@ -79,7 +79,7 @@ jobs:
     # https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability
     strategy:
       matrix:
-        msrv: [1.65.0] # let .. else .. is stable in 1.65.0
```
CyberHoward, line 82: This is a big jump, what prompted it? The comment should be up-to-date on why this MSRV is what it is.

Buckram123: I had this issue https://github.com/rust-lang/cargo/issues/10788
It got fixed at 1.71.0, I'll update the comment

CyberHoward (commented):
This PR is great, love the way this is going. Couple comments: 

- `cw-orch-environment` contains more than just the environment. Maybe it should be `cw-orch-core`? 
- `cw-orch-environment` has all the network information (which frequently changes). I suggest moving this into the daemon package or a separate package instead.
- The `Daemon` and `DaemonAsync` can both be used on their own so I think it makes sense to pull them out of the `packages` dir into the repo root.

PR #153, opened 2023-07-09
Feature/local executor

Kayanski:
This PR aims at adding a mock chain for cw-orch. 
This new chain based on the cw_remote_test executor allows executing a contract logic against real chain data.
This allows especially testing against on-chain contracts


Kayanski:
This was moved to https://github.com/AbstractSDK/cw-orch-interchain

PR #158, opened 2023-07-13
Fix doc builds

Buckram123:
Fixed few warnings, but was unable to reproduce [the error](https://docs.rs/crate/cw-orch/0.13.3/builds/854107) with `cargo doc` or docs.rs [build subcommand](https://github.com/rust-lang/docs.rs/blob/master/README.md#build-subcommand)


CyberHoward:
All checks pass so LGTM

PR #159, opened 2023-07-14
Feature/Downcast CwOrchError

Buckram123:
This PR aims to add a downcast function for `CwOrchError`
This allows comparing errors in tests


CyberHoward (approved):
Hmm yeah preferably we wouldn't have an anyerror variant but cw-multi-test uses it so we have to as well...

PR #169, opened 2023-07-27
Fix/ Remove RefCell from osmosis-test-tube sender

Buckram123:


cw-orch/src/osmosis_test_tube/core.rs
```diff
@@ -75,12 +75,8 @@ impl<S: StateInterface> OsmosisTestTube<S> {
...
         &self,
         amount: Vec<cosmwasm_std::Coin>,
-    ) -> Result<Rc<RefCell<SigningAccount>>, CwOrchError> {
-        self.app
-            .borrow()
-            .init_account(&amount)
-            .map_err(Into::into)
-            .map(|a| Rc::new(RefCell::new(a)))
+    ) -> Result<SigningAccount, CwOrchError> {
+        self.app.borrow().init_account(&amount).map_err(Into::into)
```
adairrr, line 79: Let's return the `Rc::new(a)` here so that the user doesn't have to wrap it manually when using `call_as`.

adairrr (approved):
Nice!

cw-orch/src/osmosis_test_tube/core.rs
Buckram123: Done, updated init_accounts as well, so users can do stuff like `app.set_sender(&accounts[0])` instead of `app.set_sender(&Rc::new(accounts[0]))`

Kayanski (approved):

PR #170, opened 2023-07-31
Added an install readme

Kayanski:
This PR aims at providing users with more specific guidelines as to what packages they need to install and use cw-orch.

This tutorial includes only Ubuntu specific explanation. Please feel free to add other OS tutorials


Buckram123:
@Kayanski What do you think about packing dependencies install into one command?

Kayanski:
> @Kayanski What do you think about packing dependencies install into one command?

Sure !

PR #172, opened 2023-08-01
Update/gas estimation

Kayanski:
This PR aims at having a better gas estimation accounting for low gas transactions that need a higher buffer for passing


cw-orch/src/daemon/tx_builder.rs
```diff
@@ -98,7 +100,11 @@ impl TxBuilder {
                 .await?;
             log::debug!("Simulated gas needed {:?}", sim_gas_used);
 
-            let gas_expected = sim_gas_used as f64 * GAS_BUFFER;
+            let gas_expected = if sim_gas_used < BUFFER_CHANGE_LIMIT {
+                sim_gas_used as f64 * SMALL_GAS_BUFFER
+            } else {
+                sim_gas_used as f64 * GAS_BUFFER
+            };
```
Buckram123, line 107: Nice

Buckram123 (approved):
LGTM

cw-orch/src/daemon/tx_builder.rs
```diff
@@ -10,6 +10,8 @@ use secp256k1::All;
 use super::{sender::Sender, DaemonError};
 
 const GAS_BUFFER: f64 = 1.2;
+const BUFFER_CHANGE_LIMIT: u64 = 100_000;
```
Buckram123, line 13: :skull_and_crossbones: just got 
```
Error: tx failed: out of gas in location: wasm contract; gasWanted: 117885, gasUsed: 118671: out of gas with code 11
```
Should we increase it to 150/200k?

Buckram123 (changes requested):

cw-orch/src/daemon/tx_builder.rs
Kayanski: Let's try 150k yeah

Kayanski: Done,

Buckram123 (approved):
150 worked for me, hopefully we won't have 20 buffer ranges at the end

CyberHoward:
bumped the threshold as discussed internally. Also bumped the base gas multiplier based on the chain-registry suggested gas multipliers. 

Most chains have 1.2 but some significant ones have 1.3. Since this is mainly a development and deployment tool I think reliability has much greater value than saving some cents here and there.

PR #174, opened 2023-08-04
Feature/custom mock app

### LINKED ISSUE #173: Add support for custom Apps on Mock
Currently we only support BasicApp, but some chains will have custom app implementation and supporting custom apps would help to mock test those.
Example: https://github.com/Team-Kujira/kujira-rs-testing/blob/71d30e0b17d19130595e6be4df677ff47cdf9d54/src/mock.rs#L23-L24

Buckram123:
Closes #173.
- [x] Implement CustomMock that can hold custom messages and queries. 
- [ ] Do some testing with CustomMock that contains custom messages and queries
- [ ] Add documentation on CustomMock

### DIFF
--- a/contracts/counter/tests/integration_tests.rs
+++ b/contracts/counter/tests/integration_tests.rs
@@ -15,7 +15,7 @@ const ADMIN: &str = "admin";
 // ANCHOR: integration_test
 // ANCHOR: setup
 /// Instantiate the contract in any CosmWasm environment
-fn setup<Chain: CwEnv>(chain: Chain) -> CounterContract<Chain> {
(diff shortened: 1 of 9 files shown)


CyberHoward:
Does this work?

Buckram123:
> Does this work?

Kinda, but it still doesn't solve the main issue for cases where MOST OF CONTRACTS will have `Empty` custom execute and custom query. May help to mock test contracts that have custom execute and custom query handles if they rely only on same type of handles

Buckram123:
Custom messages contracts will still have issues with ones that have different or `Empty` type app

PR #179, opened 2023-08-08
Fix/CI rust cache

Buckram123:


Buckram123:
It fixes original issue on installing rustup, but fails to mismatched types error on `mdbook-keeper` side. https://github.com/tfpk/mdbook-keeper/issues/12

adairrr (approved):

PR #187, opened 2023-08-09
Started test environment specific documentation

Kayanski:
This PR aims at introducing better docs for each environment that cw-orchestrator supports.

This also aims at having better visibility for those integrations in the docs to pique the users interests.

Fixes ORC-3


contracts/counter/README.md
```diff
@@ -0,0 +1,255 @@
...
+
+
+
+
+## Other cw-orch traits
+
+### Deploy
+
+cw-orchestrator defines the [Deploy](../../cw-orch/src/deploy.rs) trait that allows developers to create an interface that allows themselves or other developers to : 
+1. Simply deploy their codebase on a test chain or elsewhere. [See it live here](https://github.com/AbstractSDK/abstract/blob/main/framework/scripts/src/bin/full_deploy.rs#L29).
```
MbBrainz, line 182:
```suggestion
1. Easily deploy their codebase on a test chain or elsewhere. [See it live here](https://github.com/AbstractSDK/abstract/blob/main/framework/scripts/src/bin/full_deploy.rs#L29).
```

contracts/counter/README.md
```diff
@@ -0,0 +1,255 @@
...
+2. Provide bindings (addresses, code_ids), for other developers to integrate with the platform. [See it live here](https://github.com/AbstractSDK/abstract/blob/main/modules/contracts/apps/croncat/examples/deploy.rs#L32).
+
+This is even more powerful than simply exposing the single contract interfaces, because it allows full customization of the deployment procedure. It also allows shipping other custom methods for simplifying the life of the integrator.
```
MbBrainz, line 185:
```suggestion
This is more powerful than simply exposing the single contract interfaces, because it allows full customization of the deployment procedure. It also allows shipping other custom methods for simplifying the life of the integrator.
```

contracts/counter/README.md
```diff
@@ -0,0 +1,255 @@
...
+    daodao.set_account_id(dao_id)?;
+
+    let my_custom_gov_admin = GovAdmin::new("gov-admin", chain)?;
+    my_custom_gov_admin.upload()?;
+    my_custom_gov_admin.instantiate(&InstantiateMsg{account_id: dao_id});
+
+    daodao.nominate_admin(my_custom_gov_admin.address().to_string())?;
+}
+```
+In our example, all the daodao core addresses are shipped with the daodao crate directly. Projects need to verify certain conditions in order to ship this state with the deploy bundle.
```
MbBrainz, line 211: This first line is not entirely clear. I would add a little elaboration on the definition of "daodao core addresses are shipped with the daodao crate directly."

What does it mean to "ship daodao core addresses"?  the addresses of the contracts deployed according to the latest crate or the addresses of what?

contracts/counter/README.md
```diff
@@ -0,0 +1,255 @@
...
+```
+
+The `Deploy::set_contracts_state` doesn't need to be re-implemented and it allows to override the daemon state file and use the deployed addresses and code_ids instead. Doing this decouples the local state of the users from the deployed state of the maintainers/project. 
+
+You can customize the `Deploy::deployed_state_file_path` and `Deploy::load_from` methods, be we recommend doing something similar to simplify exporting the state correctly. 
+
+
+> **_NOTE_** You should check out the [Abstract bundle implementation](https://github.com/AbstractSDK/abstract/blob/main/framework/packages/abstract-interface/src/deployment.rs) to understand how to ship a bundle. In abstract, we have a file named `daemon_state.json` in the crate root with all the abstract state. We refer to it by referring to the absolute crate path and adding `"daemon_state.json"` to it. 
+> 
+> We recommend defining this location from the absolute crate path `env!("CARGO_MANIFEST_DIR")` for it to be accessible even when imported from a crate.
```
MbBrainz, line 248: Maybe add an example repo tree here

CyberHoward:
Doing some cleanup right now on this PR.

CyberHoward:
`integrations/index.md` could use a bit of text but besides that it's ok to merge!

CyberHoward (approved):

PR #194, opened 2023-08-17
RPC Implementation

Kayanski:
This PR aims at introducing RPC support into cw-orchestrator.

It allows both rpc and grpc interactions. These 2 features are feature flagged and not compatible.

Closes ORC-15


Kayanski:
This still needs to be tested and updated. Closing because this is not work that is planned in the near future

PR #195, opened 2023-08-18
Bump dep versions and move live_mock

CyberHoward:
Bumps the dependencies on ibc-relayer and ibc-registry as well as the cryptographic ed25519-dalek library.


CyberHoward:
@Kayanski Would you be able to remove the `node-tests` feature from `cw-orch` and move it to `cw-orch-daemon`, along with the tests?

Kayanski:
> @Kayanski Would you be able to remove the `node-tests` feature from `cw-orch` and move it to `cw-orch-daemon`, along with the tests?

Yes, done

PR #197, opened 2023-08-21
Update/blanket rc stateinterface

Buckram123:
DaemonState doesn't use `RefCell`. So let's implement blanket `Rc<StateInterface>` as well to avoid using newtypes or other wrappers

### DIFF
--- a/packages/cw-orch-core/src/environment/state.rs
+++ b/packages/cw-orch-core/src/environment/state.rs
@@ -77,3 +77,33 @@ impl<S: StateInterface> StateInterface for Rc<RefCell<S>> {
         (**self).borrow().deploy_details()
     }
 }
+
+impl<S: StateInterface> StateInterface for Rc<S> {
+    fn get_address(&self, contract_id: &str) -> Result<Addr, CwEnvError> {
+        (**self).get_address(contract_id)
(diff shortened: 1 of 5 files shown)


packages/cw-orch-core/src/environment/state.rs
```diff
@@ -77,3 +77,33 @@ impl<S: StateInterface> StateInterface for Rc<RefCell<S>> {
...
     }
 }
+
+impl<S: StateInterface> StateInterface for Rc<S> {
+    fn get_address(&self, contract_id: &str) -> Result<Addr, CwEnvError> {
+        (**self).get_address(contract_id)
+    }
+
+    fn set_address(&mut self, contract_id: &str, address: &Addr) {
+        (*Rc::make_mut(self)).set_address(contract_id, address)
```
CyberHoward, line 87: Just making a note here that `make_mut` will clone the inner value and set the address through that.

CyberHoward (approved):

packages/cw-orch-core/src/environment/state.rs
Buckram123: Only if we create another rc to that state, right?

PR #198, opened 2023-08-21
Minor changes + update publish script

Kayanski:
This PR aims at creating a proper release flow for cw-orch from 0.15.0 on


cw-orch/src/osmosis_test_tube/core.rs
```diff
@@ -73,6 +74,12 @@ pub struct OsmosisTestTube<S: StateInterface = MockState> {
     pub app: Rc<RefCell<OsmosisTestApp>>,
 }
 
+/// Maps a RunnerError (from osmosis-test-tube), to a CwOrch Error
+/// As we don't have a custom Error type for this integration, we use this helper function to map errors instead
+pub fn map_err(e: RunnerError) -> CwOrchError {
```
CyberHoward, line 79: Shouldn't the error type exposed by `cw-orch` be in cw-orch itself instead of using the error type in cw-orch-core? That allows us to have a `#[from(...)]` variant on the error that's feature flagged and lets us remove this code.

Kayanski: Yeah, let's go

PR #206, opened 2023-09-03
Make broadcast function public

Kayanski:
This PR aims at making the boadcast function of the sender public for using it if necessary by the users. This is not critical nor needed by any projects we work with, just thought it could be helpful


Kayanski:
Already fixed in https://github.com/AbstractSDK/cw-orchestrator/pull/200

PR #210, opened 2023-09-04
Update/tx retry strategies

Kayanski:
This PR aims at allowing transactions to not immediately fail if there is an account sequence error.
I also renderez tx retry more modular for future implementations

### DIFF
--- a/cw-orch-daemon/Cargo.toml
+++ b/cw-orch-daemon/Cargo.toml
@@ -37,6 +37,7 @@ schemars = "0.8.10"
 log = { workspace = true }
 serde_json = { workspace = true }
 thiserror = { workspace = true }
+async-recursion = "1.0.5"
 
 # Daemon deps
 sha256 = { workspace = true }
--- a/cw-orch-daemon/src/tx_broadcaster.rs
+++ b/cw-orch-daemon/src/tx_broadcaster.rs
@@ -1,0 +1,226 @@
+use std::time::Duration;
+
+use cosmrs::proto::cosmos::base::abci::v1beta1::TxResponse;
+use secp256k1::All;
+
+use crate::{
+    queriers::{DaemonQuerier, Node},
+    sender::Sender,
+    DaemonError, TxBuilder,
+};
+
+pub type StrategyAction =
+    fn(&mut TxBuilder, &Result<TxResponse, DaemonError>) -> Result<(), DaemonError>;
+
+pub struct RetryStrategy {
+    /// This correspond to errors caught when submitting transactions
+    pub broadcast_condition: fn(&TxResponse) -> bool,
+    /// This correspond to errors caught when simulating the transaction
+    pub simulation_condition: fn(&DaemonError) -> bool,
+    pub action: Option<StrategyAction>,
+    pub max_retries: BroadcastRetry,
+    pub current_retries: u64,
+    pub reason: String,
+}
+
+impl RetryStrategy {
+    pub fn new(
+        broadcast_condition: fn(&TxResponse) -> bool,
+        simulation_condition: fn(&DaemonError) -> bool,
+        action: Option<StrategyAction>,
+        max_retries: BroadcastRetry,
+        reason: String,
+    ) -> Self {
+        RetryStrategy {
+            broadcast_condition,
+            simulation_condition,
+            action,
+            max_retries,
+            current_retries: 0,
+            reason,
+        }
+    }
+}
+
+#[derive(Default)]
+pub struct TxBroadcaster {
+    strategies: Vec<RetryStrategy>,
+}
+
+pub enum BroadcastRetry {
+    Infinite,
+    Finite(u64),
+}
+
+impl TxBroadcaster {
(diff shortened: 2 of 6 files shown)


cw-orch-daemon/src/tx_broadcaster.rs
```diff
@@ -0,0 +1,226 @@
...
+    fn(&mut TxBuilder, &Result<TxResponse, DaemonError>) -> Result<(), DaemonError>;
+
+pub struct RetryStrategy {
+    /// This correspond to errors caught when submitting transactions
+    pub broadcast_condition: fn(&TxResponse) -> bool,
+    /// This correspond to errors caught when simulating the transaction
+    pub simulation_condition: fn(&DaemonError) -> bool,
+    pub action: Option<StrategyAction>,
+    pub max_retries: BroadcastRetry,
+    pub current_retries: u64,
```
CyberHoward, line 22: `(pub crate)` probably suffices?

We should try to keep as much as possible private, that's good API practice.

cw-orch-daemon/src/tx_broadcaster.rs
```diff
@@ -0,0 +1,226 @@
...
+
+pub type StrategyAction =
+    fn(&mut TxBuilder, &Result<TxResponse, DaemonError>) -> Result<(), DaemonError>;
+
+pub struct RetryStrategy {
+    /// This correspond to errors caught when submitting transactions
+    pub broadcast_condition: fn(&TxResponse) -> bool,
+    /// This correspond to errors caught when simulating the transaction
+    pub simulation_condition: fn(&DaemonError) -> bool,
+    pub action: Option<StrategyAction>,
```
CyberHoward, line 20: What's this action? Could you doc-comment it?

cw-orch-daemon/src/tx_broadcaster.rs
```diff
@@ -0,0 +1,226 @@
...
+    queriers::{DaemonQuerier, Node},
+    sender::Sender,
+    DaemonError, TxBuilder,
+};
+
+pub type StrategyAction =
+    fn(&mut TxBuilder, &Result<TxResponse, DaemonError>) -> Result<(), DaemonError>;
+
+pub struct RetryStrategy {
+    /// This correspond to errors caught when submitting transactions
```
CyberHoward, line 16:
```suggestion
    /// This function is triggerd when a submitted transaction fails.
    /// TODO: What's the returned bool used for?
```

cw-orch-daemon/src/tx_broadcaster.rs
```diff
@@ -0,0 +1,226 @@
...
+    Finite(u64),
+}
+
+impl TxBroadcaster {
+    pub fn add_strategy(mut self, s: RetryStrategy) -> Self {
+        self.strategies.push(s);
+        self
+    }
+
+    // We can't make async recursions easily because wallet is not `Sync`, so we have this shit function structure
```
CyberHoward, line 61: Would prefer to not call our own code "shit".

cw-orch-daemon/src/tx_broadcaster.rs
```diff
@@ -0,0 +1,226 @@
...
+
+        // We try and broadcast once
+        let mut tx_response = broadcast_helper(&mut tx_builder, wallet).await;
+
+        while tx_retry {
+            tx_retry = false;
+
+            // On error, we verify it matches the retry strategies. If it does, we update the retry counts and start over
+            for s in self.strategies.iter_mut() {
+                if strategy_condition_met(s, &tx_response) && can_retry(s) {
```
CyberHoward, line 77: If this condition is met for multiple strategies then the tx can be re-broadcasted multiple times for one failure, is that the desired behavior?

cw-orch-daemon/Cargo.toml
```diff
@@ -37,6 +37,7 @@ schemars = "0.8.10"
 log = { workspace = true }
 serde_json = { workspace = true }
 thiserror = { workspace = true }
+async-recursion = "1.0.5"
```
CyberHoward, line 40: I don't see this being used anywhere?

cw-orch-daemon/src/tx_broadcaster.rs
Kayanski: Nope, this is not true. The `continue` keyword ensures that it's retried only once (L96)

cw-orch-daemon/Cargo.toml
Kayanski: Because it's not used, removing

cw-orch-daemon/src/tx_broadcaster.rs
CyberHoward: Ahh okey so if the first strategy matches it only does that one? Should be documented on the constructor because that means that ordering matters.

Kayanski: No if the strategy matches the first one, it tries and correct that one and then retries. If it then matches the second one, it retries with the second strategy in mind and so forth. Anyway, Node error messages only come one per one

PR #213, opened 2023-09-05
Multiple small updates to prepare for Interchain

Kayanski:
This PR aims at implementing small changes to prepare for IBC functionalities

- Added IBC `denom-hash` function
- Added IBC `connection-end` query function
- Added Node `find_tx_by_events` query function
- commit_any function in the TxHandler trait to execute `Any` messages on environments. Compatible with Daemon and OsmosisTestTube (No mock support yet)
- Added helper to modify the `chain_id` of Mock more easily (good for identifying chains across IBC)
- Added utilization of minimum gas price and average gas price for computing gas price (less constraints on the chain objects)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 ## Unpublished
 
 - Ability to use the ExecuteFns and QueryFns traits on Units and Unnamed enum variants by @Kayanski
+- Added Sei Network
+- Make the broadcast function public on Sender
+- Added a `call_as` helper on TxHandler Trait
+- Added optional disable of sorting of ExecuteFns and QueryFns fields by @Kayanski
+- Added automatic tx retrying in case of an account sequence error by @Kayanski
+- Made transaction retrying more modular (for adding more retry cases) by Kayanski
 - Added ibc denom query (for completeness and working with ibc token transfers)
 - Added ibc connection-end query (for completeness and working with ibc connections)
 - Added transaction by event query (mainly for querying packets in and out of the chain)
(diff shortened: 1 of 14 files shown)


packages/cw-orch-mock/src/core.rs
```diff
@@ -298,6 +310,15 @@ impl<S: StateInterface> TxHandler for Mock<S> {
...
         Ok(self.app.borrow().block_info())
     }
+
+    /// Specify Empty for the type of R if you don't know the type to unwrap the response to
+    fn commit_any<R>(
+        &self,
+        _msgs: Vec<cosmrs::Any>,
+        _memo: Option<&str>,
+    ) -> Result<Self::Response, Self::Error> {
+        panic!("Un-implemented, mock env doesn't support starship messages yet")
```
CyberHoward, line 320: `starship` -> `stargate`

I wonder if it makes sense to have this in the `TxHandler` trait because it isn't implementable in `Mock`. 

If a user writes a scripts with a generic `Chain: CwEnv`, he expects the script to work with any type that meets that requirement. With this addition that is no longer the case which is not good imo :/ 


Can it be implemented maybe through a `Stargate` trait or something similar that is only implemented for Daemon and Test-Tube for now?

CyberHoward (changes requested):

packages/cw-orch-mock/src/core.rs
Kayanski: Yes good idea ! I even created an additional `cw-orch-traits` package for keeping the core stable and having additional traits implemented as we go if needed (Bank, Staking or other tx and query capabilites)

packages/cw-orch-traits/Cargo.toml
CyberHoward, line 1: This file needs some cleanup (description, comments, unused deps)

PR #218, opened 2023-09-10
Wallet balance assertion

Kayanski:
This PR aims at adding helpers to assert wallet balance content before deploying an application. 

# Changes

- Added a WalletBalanceAssertion Trait, that asserts the balance of the Chain sender is enough to run a transaction with a specific gas cost. 
- Added a prompt to load the user balance with funds before continuing the deployment (interactive)
- Added a `deploy_on_with_balance_assertion` method to the deploy trait to opt in this balance assertion

# Possible future changes
- Might have to take gas adjustments into account (maybe be clear about them in the `GAS_TO_DEPLOY` variable)
- Might make it a default to check the balance of a user before deploy, i.e., rename 2 functions : 
  - `Deploy::deploy_on_with_balance_assertion` --> `Deploy::deploy_on`
  - `Deploy::deploy_on` --> `Deploy::_deploy_on`
- Might add the possibility to error instead of asking the user to confirm they have loaded their balance with the necessary funds


Kayanski:
I need some review on the possible changes. Today, the flow is : 
- Users have to opt-in on walllet balance assertion (maybe we prefer forcing balance assertions)
- The deploy scripts become interactive (no non-interactive mode), thus preventing the usage of this function is CI/CD

What are you guys thoughts on those behaviors ?

cw-orch-daemon/src/sync/core.rs
```diff
@@ -202,3 +202,38 @@ impl TxHandler for Daemon {
         })
     }
 }
+
+impl WalletBalanceAssertion for Daemon {
```
CyberHoward, line 206: Imo this should be implemented for Wallet (i.e. the sender), not Daemon. 

`TxHandler` has an associated type `Sender`. I think that all sender-related logic should be encapsulated in traits that are a trait bound for this associated type. I.e. 

```rust
pub trait TxHandler: ChainState + Clone {

    type Sender: Clone + Bank ;
```

Then `Bank` could include a `.balance` and `.send_funds` method that's already in other places in this repo (imo send should be removed from the mock/daemon and only be available here) 

And then the assertion can be a supertrait of this `Bank` trait. I.e. 

```rust
trait BalanceAssertion: Bank {
// logic can be implemented here
}

// Blanket implementation for anything that implements Bank
impl <T: Bank> BalanceAssertion for <T> {
 // will be empty
}
```

CyberHoward: The question then becomes where we want to translate from the gas requirement to the actual fee requirement and in my opinion that should probably be done in the deploy trait because that's really where it's important. But that means that we need to somehow expose some endpoints on either the TxHandler trait that allows us to get the gas price for a specific environment. WDYT?

CyberHoward (approved):

CyberHoward (approved):

CyberHoward (approved):

PR #227, opened 2023-09-15
Add ability to set custom state on deploy object

CyberHoward:
Add the ability to set a custom state for a deployment. This can be used for test depolyments and scripting.

Fixes ORC-37 as well


Kayanski:
How / in which case would you use that ? Today, the set_contract_state function is used directly inside the Abstract Deploy::load_from method, do you see it used there as well ? 

Or is this just for tests ?

Kayanski:
How / in which case would you use that ? Today, the set_contract_state function is used directly inside the Abstract Deploy::load_from method, do you see it used there as well ? 
Or just in test cases

Kayanski (approved):

Kayanski:
Just needs changelog ?

CyberHoward:
> How / in which case would you use that ? Today, the set_contract_state function is used directly inside the Abstract Deploy::load_from method, do you see it used there as well ? Or just in test cases

We're going to need it to do migration testing. I.e. we deploy some contracts with version 1 and then migrate them with an Abstract version 2. In that case we need to load state 1 into the abstract struct version 2. 

Maybe there's a better way for us to go about migrations?

PR #228, opened 2023-09-18
Added automatically derived docs

Buckram123:
`#![warn(missing_docs)]` shouldn't warn inside derived traits/impls

### DIFF
--- a/packages/cw-orch-fns-derive/src/execute_fns.rs
+++ b/packages/cw-orch-fns-derive/src/execute_fns.rs
@@ -4,7 +4,7 @@ use convert_case::{Case, Casing};
 use proc_macro::TokenStream;
 use proc_macro2::Span;
 use quote::{format_ident, quote};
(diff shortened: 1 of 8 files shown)


CyberHoward (approved):
Awesome!

Kayanski (approved):

packages/cw-orch-fns-derive/src/execute_fns.rs
```diff
@@ -148,7 +148,6 @@ pub fn execute_fns_derive(input: DeriveInput) -> TokenStream {
 
     let derived_trait_impl = quote!(
         #[automatically_derived]
-        #[allow(undocumented)]
```
CyberHoward: Why did you remove this?

Buckram123: `automatically_derived` covers it for impl blocks!

Buckram123: Actually it won't warn on trait impls, since docs of it should be inside the trait itself

PR #233, opened 2023-09-20
No-panicking parse_network

Buckram123:
Would like to have result instead of panic


Kayanski:
I thought this was already done ! Haha, weird

PR #235, opened 2023-09-20
Added bank + fee estimation traits

Kayanski:
This PR aims at creating traits on the TxHandlers for having a common interface on all to submit transactions with scripts

### DIFF
--- a/cw-orch-daemon/src/error.rs
+++ b/cw-orch-daemon/src/error.rs
@@ -10,6 +10,8 @@ pub enum DaemonError {
     #[error("JSON Conversion Error")]
     SerdeJson(#[from] ::serde_json::Error),
     #[error(transparent)]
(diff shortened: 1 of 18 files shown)


CyberHoward:
is this a replacement of #218 ? Which one should be merged first?

Kayanski:
> is this a replacement of #218 ? Which one should be merged first?

This was only a test implementation, but this doesn't work for now.
This will be useful when implementing module capabilites (query + execution), but cannot be used for wallet balance assertion capabilites. 
The wallet balance assertion should be merged first (and refactored to only be used with the Daemon Sender) and this one comes later

Kayanski:
This is not for now, an issue has been created : https://github.com/AbstractSDK/cw-orchestrator/issues/268

PR #243, opened 2023-09-28
Add find_wasm_path_with_build_postfix function to ArtifactsDir

CyberHoward:
Add the ability to provide a build post-fix that is either none, custom or network-based. This is useful in combination with the featured optimizer.cf


packages/cw-orch-core/src/contract/paths.rs
```diff
@@ -142,13 +142,33 @@ mod artifacts_dir {
...
                     let path = entry.ok()?.path();
                     let file_name = path.file_name().unwrap_or_default().to_string_lossy();
-                    if path.is_file()
-                        && path.extension().unwrap_or_default() == "wasm"
-                        && file_name.contains(name)
+                    if !path.is_file() {
+                        return None;
+                    }
+
+                    if (path.extension().unwrap_or_default() == "wasm"
```
Kayanski, line 165: This line is weird, what are the conditions we want here : 
1. We need a wasm file
2. We need it to either be non-postfixed or post-fixed

It looks like the condition here is just is_default_artifact, and that's it, don't understand why you would need the postfix validation here. You are not adding build postfix there in the path anyway

Kayanski (changes requested):

packages/cw-orch-core/src/contract/paths.rs
CyberHoward: It's different because `is_default_artifact` matches on the default name (i.e only the contract name and no postfixes) so it would return false for a build with a postfix.

PR #251, opened 2023-10-10
Changed conditional implementation for daemon + mock

Kayanski:
This PR aims at unifying the conditional upload experience across scripting / testing

In practice, we change the way those functions are defined to make it available across all CwEnvs


packages/cw-orch-core/src/environment/cosmwasm_environment.rs
```diff
@@ -88,3 +91,11 @@ pub trait TxHandler: ChainState + Clone {
         chain
     }
 }
+
+pub trait WasmCodeQuerier: CwEnv {
+    fn get_contract_hash(&self, code_id: u64) -> Result<String, <Self as TxHandler>::Error>;
```
CyberHoward, line 96: FYI: https://rust-lang.github.io/api-guidelines/naming.html#:~:text=Getter%20names%20follow%20Rust%20convention,for%20getters%20in%20Rust%20code.&text=The%20get%20naming%20is%20used,the%20content%20of%20a%20Cell%20.

Do we want to follow these API guidelines? I'd like us to.

We follow them in the TxHander:

```rust
    /// Gets the address of the current wallet used to sign transactions.
    fn sender(&self) -> Addr;

    /// Sets wallet to sign transactions.
    fn set_sender(&mut self, sender: Self::Sender);
```

packages/cw-orch-mock/src/core.rs
```diff
@@ -310,6 +310,27 @@ impl<S: StateInterface> TxHandler for Mock<S> {
     }
 }
 
+impl WasmCodeQuerier for Mock {
+    /// Returns whether the checksum of the WASM file matches the checksum of the latest uploaded code for this contract.
```
CyberHoward, line 314: comments are outdated

packages/cw-orch-mock/src/core.rs
```diff
@@ -310,6 +310,27 @@ impl<S: StateInterface> TxHandler for Mock<S> {
...
+    fn get_contract_hash(&self, code_id: u64) -> Result<String, CwEnvError> {
+        let code_info = self.app.borrow().wrap().query_wasm_code_info(code_id)?;
```
CyberHoward, line 316: Does this actually return something? Is there some code in the `App` that handles this query? If so, what does it return? Can we add that information as a comment or doc-comment on this method?

CyberHoward (changes requested):
Some minor changes, I like the implementation!

PR #257, opened 2023-10-17
Update/add full deploy

Kayanski:
This PR aims at integrating the full_deploy function inside the Deploy trait for allowing all projects to simply deploy their projects without blocking and with being able to retry if a deployment fails on a chain.

Closes ORC-43

You can find a test here : https://github.com/AbstractSDK/abstract/pull/101


packages/cw-orch-core/src/contract/deploy.rs
```diff
@@ -70,6 +79,102 @@ pub trait Deploy<Chain: CwEnv>: Sized {
         Self::store_on(chain)
     }
 
+    /// Deploys the applications on all chains indicated in `chains`.
+    /// Arguments :
+    ///  - `networks`` is a vector of :
+    ///     - Chain objects
+    ///     - Additional deploy data needed for the deployment of the structure on each platform
+    fn full_deploy(
```
CyberHoward, line 87: Can we rename this to `multi_network_deploy` or something of that vein?

PR #259, opened 2023-10-17
Update/better env variable handling

Kayanski:
This PR aims at handling the env variables better with parsing and default values being centralized in one place.


packages/cw-orch-core/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 pub mod contract;
 pub mod env;
+pub use env as CwOrchEnvVars;
```
CyberHoward, line 3: oeeehh yikes! Have you seen this pattern/naming somewhere before? I don't really like it. People will think it's a type when it isn't.

packages/cw-orch-core/src/env.rs
CyberHoward, line 1: I was thinking you were going to add values to the env enums. 

I.e. 

```rust
enum EnvVar {
    Abc(String),
    Xyz(u8),
}
```

I think this crate might be of interest: https://github.com/Peternator7/strum

If you could give it a look and still think this is good then it's also good for me!

CyberHoward: This one specifically looks interesting: https://docs.rs/strum_macros/0.25.3/strum_macros/derive.EnumDiscriminants.html

packages/cw-orch-core/src/lib.rs
Kayanski: Yeah you're right, will change

packages/cw-orch-core/src/env.rs
Kayanski: One issue why I chose this implementation is that I can't get a typed value from an enum variant.

For instance in your example, I can't have a at the same time : 
- EnvVar::Abc.value() method that gets me a String and 
- EnvVar::Xyz.value() method that gets me a u8

CyberHoward: The idea would be that you can't construct the type without it having its value. 

```rust
EnvVar {
   A(String),
   B(u16),
   ...
```

And then use a crate like this: https://docs.rs/strum_macros/0.25.3/strum_macros/derive.EnumTryAs.html
to get its value when you need it.

PR #263, opened 2023-10-23
Update/add all chains

Kayanski:
This PR aims att adding the whole chain registry (from hermes) into orchestrator/

Issue : the GRPC addresses in the registry don't always have the right format for orchestrator (example for dorahacks, chain id and grpc are wrong)


Kayanski:
This is not reliable enough and doesn't have the right parameters for querying chains.

PR #277, opened 2023-11-23
Add WasmCodeQuerier support to ForkMock

CyberHoward:
- Add `WasmCodeQuerier` to `CwEnv` trait bound
- update conditional trait's bounds

### DIFF
--- a/cw-orch/src/osmosis_test_tube/core.rs
+++ b/cw-orch/src/osmosis_test_tube/core.rs
@@ -1,5 +1,6 @@
 use crate::contract::WasmPath;
 use crate::prelude::Uploadable;
+use cw_orch_core::environment::WasmCodeQuerier;
 use cw_orch_traits::stargate::Stargate;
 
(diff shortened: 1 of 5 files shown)


cw-orch/src/osmosis_test_tube/core.rs
```diff
@@ -338,3 +339,16 @@ impl Stargate for OsmosisTestTube {
         })
     }
 }
+
+impl WasmCodeQuerier for OsmosisTestTube {
```
CyberHoward, line 343: Do you know if there's a way to implement these functions for test-tube @Kayanski?

Kayanski: yes ! With raw queries :)

Kayanski: It works well now :)

CyberHoward:
LGTM!!

Issue #96, opened 2023-04-25
Allow for default id in generated `new` function.


CyberHoward:
This code is currently generated for the `contract` macro.
```rust
        impl<Chain: ::boot_core::CwEnv> #name<Chain> {
            pub fn new(contract_id: &str, chain: Chain) -> Self {
                Self(
                    ::boot_core::Contract::new(contract_id, chain)
                )
            }
        }
```
We should have the ability to specify a default ID for the contract so we don't need to provide it on construction every time.

CyberHoward:
I'd prefer not having the version in the name, it will cause issues indexing state

PR #283, opened 2023-12-04
Added id to the interface

Buckram123:
Fixes ORC-56

As discussed in a ticket if interface contains id at the end we auto-generate `new` constructor with this "contract_id" or make users build their own constructor.

Currently supported syntax:
```rust
#[interface(InstantiateMsg, ExecuteMsg, QueryMsg, MigrateMsg, id=MY_CONTRACT)]
// or 
#[interface(InstantiateMsg, ExecuteMsg, QueryMsg, MigrateMsg, id="contract_literal")]
```
It is up for discussion if we want to change it or add more restrictions

Closes https://github.com/AbstractSDK/cw-orchestrator/issues/96

TODO:
- [ ] Update docs and comments according to the changes (after we agree on syntax)


Buckram123:
> I'd prefer using the const as the ID is something that's often used elsewhere

Right now it supports both or would you prefer for it to be only const?

packages/cw-orch-contract-derive/src/lib.rs
```diff
@@ -163,19 +192,26 @@ pub fn interface(attrs: TokenStream, input: TokenStream) -> TokenStream {
     };
 
     let name = cw_orch_struct.ident.clone();
+    let default_num = if let Some(id_expr) = default_id {
```
Kayanski, line 195: I don't like setting the id forcefully here. Can we keep the old version if no id was provided and put this new version only if a default id is provided ?

packages/cw-orch-contract-derive/src/lib.rs
```diff
@@ -163,19 +192,26 @@ pub fn interface(attrs: TokenStream, input: TokenStream) -> TokenStream {
...
+            impl <Chain: ::cw_orch::prelude::CwEnv, #all_generics> #name<Chain, #all_generics> {
+                pub fn new(chain: Chain) -> Self {
+                    Self(
+                        ::cw_orch::contract::Contract::new(#id_expr, chain)
+                    , #(#all_phantom_marker_values,)*)
+                }
+            }
+        )
+    } else {
+        quote!()
```
Kayanski, line 206: Like simply, reput the old version here

packages/cw-orch-contract-derive/src/lib.rs
Buckram123: That's actually the goal to make it not generate `new` by default so it's handled by the contract maintainer

Kayanski: Oh, hmm, ok ! Adds an extra argument that has nothing to do with messages, but that could work

Buckram123: > Oh, hmm, ok ! Adds an extra argument that has nothing to do with messages, but that could work

We totally could change syntax of it like 
```rust
#[interface(InstantiateMsg, ExecuteMsg, QueryMsg, MigrateMsg)]
#[contract_id=CONTRACT_ID]
```

PR #285, opened 2023-12-11
Update dependencies

Kayanski:
This PR aims at updating dependencies : 
- cosmrs 0.15.0 --> Bumps MSRV to 1.72
- tonic 0.10

### DIFF
--- a/packages/cw-orch-traits/Cargo.toml
+++ b/packages/cw-orch-traits/Cargo.toml
@@ -13,5 +13,6 @@ readme = "README.md"
 
 [dependencies]
 cw-orch-core = { workspace = true }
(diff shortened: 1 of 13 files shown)


packages/cw-orch-traits/Cargo.toml
```diff
@@ -13,5 +13,5 @@ readme = "README.md"
 
 [dependencies]
 cw-orch-core = { workspace = true }
-prost-types = "0.11.0"
-prost = "0.11.0"
+prost-types = { version = "0.11.9" }
```
CyberHoward, line 16: use workspace here as well?

Kayanski: No, because it's not compatible with osmosis-test-tube. Will add a comment there

PR #293, opened 2023-12-20
Update/injective test tube

Kayanski:
This PR aims at adding Injective Test Tube to cw-orch.

This is not prioritary, but this is tested


mindbend0x:
Any particular reason why this PR wasn't merged? Seems like the work was already done. 

Asking because I'm currently running into issues with osmosis-test-tube being outdated.

Kayanski:
It was hard to maintain, as the code was not maintained properly, with breaking changes in minor versions. 
So we decided not to maintain that inside the main branch.
Right now, we are not actively maintaining the tool, as the Abstract organization is on pause at the moment.

mindbend0x:
Got it, thank you @Kayanski

PR #295, opened 2024-01-08
Fix broken state file path resolution

Kayanski:
This PR aims at resolving state file path resolution 
Closes ORC-46

# Discussion

Today we have 2 variables to set the state file path : 
- CW_ORCH_STATE_FOLDER for the folder of the state file
- STATE_FILE for the file name

This is too much. I propose to remove the first one and keep only the latter and follow the rules : 
`folder/file.json` is `.cw-orchestrator/folder/file.json`

`./folder/file.json` is `call_dir/folder/file.json`

`/usr/var/file.json` is `/usr/var/file.json`


Buckram123:
> keep only the latter and follow the rules : `folder/file.json` is `.cw-orchestrator/folder/file.json`
> 
> `./folder/file.json` is `call_dir/folder/file.json`

Not sure how I feel about it. For me it looks messier than ability to set `CW_ORCH_STATE_FOLDER`, the only rule I would keep is: absolute path is absolute path

Kayanski:
> The weirdness comes at the point that we decide to append cw-orch path without any additional input. What if we use something like `shellexpand::full` for this and set default CW_ORCH_STATE_FOLDER env instead So the rules would be:
> 
>     * `folder/file.json` is `folder/file.json`
> 
>     * `$CW_ORCH_STATE_FOLDER/folder/file.json` is `~/.cw-orchestrator/folder/file.json`
> 
>     * `$USR_ENV/file.json` is `/foo/bar/file.json`
> 
>     * `./folder/file.json` is `./folder/file.json`
> 
>     * `/usr/var/file.json` is `/usr/var/file.json`
> 
> 
> @Kayanski thoughts

We chose to use .cw-orchestrator by default to make sure that people don't need to be in a specific folder to run their scripts.
For instance, when execution `cargo run --bin daemon_test` in the root or in the cw-orch folder would produce 2 different files which is VERY bad behavior.
So we decided that using a ~/.cw-orchestrator folder was a good solution to this issue.

With `./` specific behavior, we still wanted the user to use files in the current dir if they really want it and without "./" to be able to use the non-error behavior

Now, we need to decide what to do for paths. Using $VARIABLE inside env files is very error-prone no ?

cw-orch-daemon/src/state.rs
```diff
@@ -131,6 +112,37 @@ impl DaemonState {
...
+                let actual_relative_path = env_file_path.strip_prefix("./")?;
+                current_dir.join(actual_relative_path)
+            } else {
+                let state_folder = default_state_folder()?;
+
+                // We need to create the default state folder if it doesn't exist
+                std::fs::create_dir_all(state_folder.clone())?;
+
+                state_folder.join(env_file_path)
+            }
```
Buckram123, line 136: Wonder how it would handle `../`

Buckram123 (approved):
Not sure how `../` would be handled. Agree it's much cleaner!

PR #298, opened 2024-01-09
Separate Querier and TxHandler

Kayanski:
This PR aims at separating the querying and the transaction signing capabilities of CwEnv.

Closes ORC-61

## Changes

1. Removed BankQuerier : Replaced by BankQuerierGetter and BankQuerier on top of it
2. Removed WasmCodeQuerier: Replaced by WasmQuerierGetter and WasmQuerier. Local hash function is also added in this trait


CyberHoward:
> ## Deprecation issues
> The block_info function of TxHandler is now marked as deprecated but we can't implement the exact same function on the QueryHandler trait because it would conflict with this function as long as it exists.
> 
> I think that we need to publish a breaking version for this change because of that !

Traits can have the same function. When calling the function the user needs to have either of the traits imported. If they are both imported they need to specify which trait they want to use. i.e. 

`TxHandler::block_info(chain)`

So no breaking change required here afaik

Kayanski:
> > ## Deprecation issues
> > The block_info function of TxHandler is now marked as deprecated but we can't implement the exact same function on the QueryHandler trait because it would conflict with this function as long as it exists.
> > I think that we need to publish a breaking version for this change because of that !
> 
> Traits can have the same function. When calling the function the user needs to have either of the traits imported. If they are both imported they need to specify which trait they want to use. i.e.
> 
> `TxHandler::block_info(chain)`
> 
> So no breaking change required here afaik

This Error that show up at the following line shows exactly what I meant with this comment : https://github.com/AbstractSDK/cw-orchestrator/blob/update/separate-query-tx/contracts/counter/src/interface.rs#L48

CyberHoward:
Can we maybe start working with milestones in github for breaking versions? I created one for `0.20`

Have you merged anything that was breaking?

Kayanski:
> Can we maybe start working with milestones in github for breaking versions? I created one for `0.20`
--> OK

> Have you merged anything that was breaking?
--> No

Kayanski:
In standby until https://github.com/AbstractSDK/cw-orchestrator/pull/318 is done

packages/cw-orch-mock/src/core.rs
```diff
@@ -281,34 +281,53 @@ impl<S: StateInterface> TxHandler for Mock<S> {
...
-            self.sender.clone(),
-            init_msg,
-            coins,
-            label.unwrap_or("contract_init"),
-            admin.map(|a| a.to_string()),
-        )?;
-        // add contract address to events manually
-        let mut event = Event::new("instantiate");
-        event = event.add_attribute("_contract_address", addr);
+            label: label.unwrap_or("contract_init").to_string(),
```
CyberHoward, line 287: Didn't we have a nicer lable?

packages/cw-orch-core/src/environment/queriers/wasm.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use crate::{
+    contract::interface_traits::{ContractInstance, Uploadable},
+    environment::TxHandler,
+    CwEnvError,
+};
+
+use super::{Querier, QueryHandler};
+
+pub trait WasmQuerierGetter<E> {
```
CyberHoward, line 12: Should be able to remove this getter

packages/cw-orch-core/src/environment/queriers/node.rs
```diff
@@ -0,0 +1,31 @@
+use cosmwasm_std::BlockInfo;
+
+use crate::environment::IndexResponse;
+use std::fmt::Debug;
+
+use super::Querier;
+
+pub trait NodeQuerierGetter<E> {
```
CyberHoward, line 8: As well as this one

cw-orch-daemon/src/queriers/staking.rs
```diff
@@ -1,23 +1,33 @@
...
     }
 }
 
+impl Querier for Staking {
+    type Error = DaemonError;
+}
+
+impl QuerierGetter<Staking> for Daemon {
+    fn querier(&self) -> Staking {
+        Staking::new_async(self.channel())
```
CyberHoward, line 24: Can it not get the runtime handle from the daemon object?

CyberHoward (changes requested):
Think some code can be removed and daemon queriers should be constructed with runtime by default

packages/cw-orch-mock/src/core.rs
Kayanski: Not for the mock package I don't think

cw-orch-daemon/src/queriers/staking.rs
Kayanski: But it can. Keeping the same structure as for Queriers with traits

CyberHoward (approved):

PR #300, opened 2024-01-10
Remove/cw multi test from core

Kayanski:
This PR aims at removing cw-multi-test from cw-orch-core to stabilize the core as much as possible.

This required to create our own contract structure and trait that we need to maintain (https://github.com/AbstractSDK/cw-orchestrator/pull/300/files#diff-fed029117aa3b723f79d7a391150cd435a4e2176f64f73295181cd731cdefafc)


packages/cw-orch-core/src/contract/contract_wrapper.rs
CyberHoward, line 1: Isn't this something we could ask Confio to publish in a separate package instead of cloning it? 

How much work would it be to do exactly that and create a PR on cw-plus?

Kayanski: Should be fast

CyberHoward: Could we give that a shot and see what they have to say about it?

Kayanski:
Closing this as Confio has chosen to stabilize cw-multi-test instead of having a separate crate of contracts

PR #302, opened 2024-01-12
Add more examples

Kayanski:
This PR aims at adding more examples and doc examples for users to interact with.

# Issue with async daemon
It's difficult to interact with the async daemon and contracts
--> For instance, the upload needs an Uploadable argument, but it's not possible to create one with the AsyncDaemon because it doesn't implement TxHandler.  
Should we remove that example or change AsyncDaemon a little ? 
I don't think it's worth the effort to be honest


CyberHoward (approved):
This is great! For the `AsyncDaemon` we can leave it as it is. 

If we really need a good async implementation in the future we can create an `AsyncTxHandler` trait that's equivalent to the current `TxHandler` but is all async. But that's not worth doing at the moment.

cw-orch/Cargo.toml
```diff
@@ -20,12 +20,20 @@ all-features = true
...
+
+[[example]]
+name = "testnet_daemon"
 required-features = ["daemon"]
 
 [[example]]
 name = "osmosis_test_tube"
 required-features = ["osmosis-test-tube"]
+[[example]]
+
```
CyberHoward, line 34:
```suggestion
```

PR #305, opened 2024-01-12
Feature/add authz wrapping

Kayanski:
This PR aims at introducing authz support for cw-orch

Closes ORC-19
# Usage

You can use 

```rust
let daemon = Daemon::builder()
    .chain(networks::LOCAL_JUNO)
    .handle(runtime.handle())
    .authz_granter(granter.clone())
    .build()?
```
or
```rust
daemon.with_authz(granter)
```

# TODO
- [ ] Documentation


cw-orch-daemon/src/sender.rs
```diff
@@ -48,14 +48,29 @@ pub type Wallet = Rc<Sender<All>>;
...
     pub private_key: PrivateKey,
     pub secp: Secp256k1<C>,
     pub(crate) daemon_state: Rc<DaemonState>,
+    pub(crate) options: SenderOptions,
+}
+
+#[derive(Default, Clone)]
+#[non_exhaustive]
+pub struct SenderOptions {
+    pub authz_granter: Option<String>,
```
Buckram123, line 62: It doesn't have a constructor method, but marked with `non_exhaustive`, which makes impossible creation of desired struct
Using this workaround rn:
```rust
let sender_options = {
    let mut sender_options = SenderOptions::default();
    sender_options.authz_granter = Some(granter);
    sender_options
};
```

cw-orch-daemon/src/builder.rs
```diff
@@ -53,6 +55,12 @@ impl DaemonAsyncBuilder {
         self
     }
 
+    /// Specifies wether authz should be used with this daemon
```
adairrr, line 58:
```suggestion
    /// Specifies whether authz should be used with this daemon
```

cw-orch-daemon/src/builder.rs
```diff
@@ -53,6 +55,12 @@ impl DaemonAsyncBuilder {
...
+    pub fn with_authz(&mut self, granter: impl ToString) -> &mut Self {
```
adairrr, line 59: Would prefer `with_authz_granter` for clarity. `with_authz` is ambiguous as to what the functionality would be.

cw-orch-daemon/src/sender.rs
```diff
@@ -92,6 +117,10 @@ impl Sender<All> {
         Ok(sender)
     }
 
+    pub fn with_authz(&mut self, granter: impl Into<String>) {
```
adairrr, line 120: Same comment here on the naming.

cw-orch-daemon/src/sender.rs
```diff
@@ -115,13 +144,21 @@ impl Sender<All> {
         Ok(self.pub_addr()?.to_string())
     }
 
+    pub fn message_sender(&self) -> Result<AccountId, DaemonError> {
```
adairrr, line 147: We should add a doc comment here, and wouldn't `msg_sender` be more consistent with use of `msg` elsewhere?

cw-orch-daemon/tests/authz.rs
```diff
@@ -0,0 +1,111 @@
+mod common;
+#[cfg(feature = "node-tests")]
+mod tests {
+    /*
+        DaemonAsync contract general tests
+    */
```
adairrr, line 6:
```suggestion
    /*
        Authz tests
    */
```

cw-orch-daemon/src/builder.rs
adairrr: Also - the other builders don't use the `with_` prefix, is there a reason for choosing to use it for authz specifically?

cw-orch-daemon/tests/authz.rs
```diff
@@ -0,0 +1,111 @@
...
+                .encode_to_vec(),
+            }],
+            None,
+        )?;
+
+        // Check Query of the grant
+        let authz_querier: Authz = daemon.query_client();
+        let grants: QueryGrantsResponse = runtime.handle().block_on(async {
+            authz_querier
+                .grants(sender.clone(), grantee.clone(), MsgSend::type_url(), None)
```
adairrr, line 81: Might as well check grantee grants and granter grants as well.

cw-orch-daemon/src/builder.rs
Kayanski: Yes, changing

cw-orch-daemon/src/sync/core.rs
CyberHoward, line 1: Can these `with_xyz` functions be rename to `set_xyz`?

Kayanski: It's not really a setter though, because it returns a new instance !

CyberHoward: Ah yeah. Is this the most useful API? It's a pattern I haven't worked with before. 

Anyway I think the `with_xyz` is a bit misleading as it returns a new object. Maybe `clone_with_xyz` would be better? 

But these functions also don't allow to easily set multiple things, like both authz and feegrant. It does it but it clones everything multiple times. 

alternative with setter: 

```
let mut authz_daemon = daemon.clone()
authz_daemon.set_xyz().set_klm();
```

cw-orch-daemon/src/sender.rs
```diff
@@ -61,6 +61,8 @@ pub struct Sender<C: Signing + Context> {
 pub struct SenderOptions {
     pub authz_granter: Option<String>,
     pub fee_granter: Option<String>,
+    /// Wallet mnemonic
+    pub mnemonic: Option<String>,
```
CyberHoward, line 65: Do we need this here? A bit confusing to have both a `Sender` and a mnemonic

Kayanski: Yes. It's done to satisfy 2 use cases : 
- Building a daemon with a custom mnemonic
- Allowing a daemon to be `rebuilt` from an existing sender

One solution to avoid this could be to construct the sender as soon as the mnemonic is given but that goes against the builder principle.
Another one would be to be able to extract all senderOptions from the built sender (which is not possible today because the mnemonic for instance is not stored)

CyberHoward: How about this, we move `mnemonic` out of `SenderOptions` and create an enum 

```rust
enum Wallet {
  Sender(Sender),
  Mnemonic(String),
}
```

And have a field in the builder with this

```rust
struct Builder {
  wallet: Option<Wallet>
}
```

When re-building this wallet is cloned (and it will be Sender type) 

If you want to set a new mnemonic you use the `.mnemonic` on the builder to overwrite the sender object.

cw-orch-daemon/src/sync/core.rs
adairrr: `with_xyz` is common practice for builder chaining, where as the `set` methods mutate the original and don't return it.

CyberHoward: we went for a different design

cw-orch-daemon/src/sender.rs
```diff
@@ -239,6 +312,20 @@ impl Sender<All> {
     ) -> Result<CosmTxResponse, DaemonError> {
         let timeout_height = Node::new(self.channel()).block_height().await? + 10u64;
 
+        let msgs = if self.options.authz_granter.is_some() {
```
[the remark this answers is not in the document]
CyberHoward: Nevermind, it's in tx_builder!

CyberHoward (approved):

PR #306, opened 2024-01-12
Nicolas/orc 67 no way to set hd index

Kayanski:
This PR aims at introducing index modifications on the sender object.
Depends on https://github.com/AbstractSDK/cw-orchestrator/pull/305


Buckram123 (approved):
LGTM

CyberHoward (approved):
Is there a reason why you need this? 

The hd index is like a derivation sequence right? So you can create multiple wallets from the same private key?

Kayanski:
> Is there a reason why you need this?
> 
> The hd index is like a derivation sequence right? So you can create multiple wallets from the same private key?

You can create multiple wallets from the same mnemonic yes ! It was a request from @Buckram123

Buckram123:
> Is there a reason why you need this?

For example this signer: https://github.com/cosmology-tech/starship/blob/5dcc918dc7d87a713bcde684293f5cdf5df1c63c/charts/devnet/scripts/neutron/create-neutron-genesis.sh#L45-L47

I have mnemonic, but have no way to sign it using cw-orch

PR #309, opened 2024-01-15
Added instantiate2

Kayanski:
This PR aims at adding instantiate2 support directly inside the TxHandler trait

### DIFF
--- a/cw-orch-daemon/src/core.rs
+++ b/cw-orch-daemon/src/core.rs
@@ -11,16 +11,18 @@ use super::{
 
 use cosmrs::{
     cosmwasm::{MsgExecuteContract, MsgInstantiateContract, MsgMigrateContract},
+    proto::cosmwasm::wasm::v1::MsgInstantiateContract2,
     tendermint::Time,
-    AccountId, Denom,
+    AccountId, Any, Denom,
 };
-use cosmwasm_std::{Addr, Coin};
+use cosmwasm_std::{Addr, Binary, Coin};
 use cw_orch_core::{
     contract::interface_traits::Uploadable,
     environment::{ChainState, IndexResponse},
     log::transaction_target,
 };
 use flate2::{write, Compression};
+use prost::Message;
 use serde::{de::DeserializeOwned, Serialize};
 use serde_json::from_str;
 use std::{
@@ -138,6 +140,44 @@ impl DaemonAsync {
         };
 
         let result = sender.commit_tx(vec![init_msg], None).await?;
+
+        log::info!(target: &transaction_target(), "Instantiation done: {:?}", result.txhash);
+
+        Ok(result)
+    }
+
+    /// Instantiate a contract.
+    pub async fn instantiate2<I: Serialize + Debug>(
+        &self,
+        code_id: u64,
+        init_msg: &I,
+        label: Option<&str>,
+        admin: Option<&Addr>,
+        coins: &[Coin],
+        salt: Binary,
+    ) -> Result<CosmTxResponse, DaemonError> {
+        let sender = &self.sender;
+
+        let init_msg = MsgInstantiateContract2 {
+            code_id,
+            label: label.unwrap_or("instantiate_contract").to_string(),
+            admin: admin.map(|a| a.to_string()).unwrap_or("".to_string()),
+            sender: sender.address()?.to_string(),
+            msg: serde_json::to_vec(&init_msg)?,
+            funds: proto_parse_cw_coins(coins)?,
+            salt: salt.to_vec(),
+            fix_msg: false,
+        };
+
+        let result = sender
+            .commit_tx_any(
+                vec![Any {
+                    type_url: "/cosmwasm.wasm.v1.MsgInstantiateContract2".to_string(),
+                    value: init_msg.encode_to_vec(),
+                }],
+                None,
+            )
+            .await?;
 
         log::info!(target: &transaction_target(), "Instantiation done: {:?}", result.txhash);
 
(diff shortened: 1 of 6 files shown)


packages/cw-orch-core/src/environment/cosmwasm_environment.rs
```diff
@@ -46,6 +46,17 @@ pub trait TxHandler: ChainState + Clone {
...
+    fn instantiate2<I: Serialize + Debug>(
+        &self,
+        code_id: u64,
+        init_msg: &I,
+        label: Option<&str>,
+        admin: Option<&Addr>,
+        coins: &[cosmwasm_std::Coin],
+        salt: Binary,
+    ) -> Result<Self::Response, Self::Error>;
+
```
Buckram123, line 59: Is it possible to return (yet) not executed instantiate2 struct instead? Main use for instantiate2 will be knowing the address, before instantiating I'm expecting experience to be something like: 
```rust
let contract1 = chain.instantiate2(...)?;
let contract1_addr = contract1.addr()?;
let contract2 = chain.instantiate2(...)?;
let contract2_addr = contract2.addr()?;
// ...
contract1.submit(Contract1InstantiateMsg{
    contract2_addr: contract2_addr.to_string()
})?;
...
```

packages/cw-orch-mock/src/core.rs
```diff
@@ -238,20 +238,50 @@ impl<S: StateInterface> TxHandler for Mock<S> {
...
+        init_msg: &I,
+        label: Option<&str>,
+        admin: Option<&Addr>,
+        coins: &[cosmwasm_std::Coin],
+        salt: Binary,
+    ) -> Result<Self::Response, CwEnvError> {
+        let msg = WasmMsg::Instantiate2 {
+            admin: admin.map(|a| a.to_string()),
+            code_id,
+            label: label.unwrap_or("contract_init").to_string(),
```
Buckram123, line 272: labels is perfect place to add some cw-orch ads, haha

packages/cw-orch-core/src/contract/contract_instance.rs
```diff
@@ -171,6 +171,56 @@ impl<Chain: TxHandler + QueryHandler + Clone> Contract<Chain> {
...
 
+    /// Initializes the contract
+    pub fn instantiate2<I: Serialize + Debug>(
+        &self,
+        msg: &I,
+        admin: Option<&Addr>,
+        coins: Option<&[Coin]>,
+    ) -> Result<TxResponse<Chain>, CwEnvError> {
+        log::info!(
+            target: &contract_target(),
```
Buckram123, line 182: it's not instantiate2

cw-orch-daemon/src/core.rs
```diff
@@ -144,6 +146,44 @@ impl DaemonAsync {
...
+        admin: Option<&Addr>,
+        coins: &[Coin],
+        salt: Binary,
+    ) -> Result<CosmTxResponse, DaemonError> {
+        let sender = &self.sender;
+
+        let init_msg = MsgInstantiateContract2 {
+            code_id,
+            label: label.unwrap_or("instantiate_contract").to_string(),
+            admin: admin.map(|a| a.to_string()).unwrap_or("".to_string()),
```
Buckram123, line 164:
```suggestion
            admin: admin.map(ToString::to_string).unwrap_or_default(),
```
small nit

cw-orch-daemon/src/core.rs
```diff
@@ -144,6 +146,44 @@ impl DaemonAsync {
...
+            sender: sender.address()?.to_string(),
+            msg: serde_json::to_vec(&init_msg)?,
+            funds: proto_parse_cw_coins(coins)?,
+            salt: salt.to_vec(),
+            fix_msg: false,
```
Buckram123, line 169: do we want fix_msg as parameter instead?

cw-orch-daemon/src/core.rs
```diff
@@ -144,6 +146,44 @@ impl DaemonAsync {
...
+        };
+
+        let result = sender
+            .commit_tx_any(
+                vec![Any {
+                    type_url: "/cosmwasm.wasm.v1.MsgInstantiateContract2".to_string(),
```
Buckram123, line 175: Are there any advantages for keeping it raw instead  MsgInstantiateContract2::TYPE_URL/MsgInstantiateContract2::type_url() (depending on a version)

Buckram123 (changes requested):
Would be nice to slightly improve user experience using instanitate2

cw-orch-daemon/src/core.rs
Buckram123: Nvm, don't think we want to use it

packages/cw-orch-core/src/environment/cosmwasm_environment.rs
Kayanski: I'm not sure I understand what you mean here ? You want a function that would return the instantiate2 addr if ever it was instantiated ?

packages/cw-orch-core/src/contract/contract_instance.rs
Kayanski: Corrected, yeah forgot to change

cw-orch-daemon/src/core.rs
Kayanski: Yes !

cw-orch-daemon/src/core.rs
Kayanski: There is no Type url definition on this one unfortunately

packages/cw-orch-core/src/environment/cosmwasm_environment.rs
Buckram123: > I'm not sure I understand what you mean here ? You want a function that would return the instantiate2 addr if ever it was instantiated ?

Yes, I want function that can return the address of pre-filled, but not executed instantiate2 message.

Buckram123: Using helpers such as https://docs.rs/cosmwasm-std/latest/cosmwasm_std/fn.instantiate2_address.html would be annoying, so I would prefer it to be as a part of API

Kayanski: OK,implemented what we discussed

CyberHoward: So this is now a method on the wasm querier correct?

CyberHoward (approved):

PR #303, opened 2024-01-12
Changed mock interface + corrected warnings

Kayanski:
This PR aims at 
- Allowing any string (or impl Into<String>) as sender for Mock
- Allowing any string (or impl Into<String>) as address in BankSetter
- Exporting cworchenvvars inside cw_orch directly

PR #312, opened 2024-01-17
Fixed mock app with bech32 enforcement

Kayanski:
TODO : Updated mock to pass tests with the same API (accepting `sender` as a valid address)
--> Modify call as as well to present a better api for users ?

Depends on https://github.com/AbstractSDK/cw-orchestrator/pull/303

### DIFF
--- a/packages/cw-orch-mock/src/core.rs
+++ b/packages/cw-orch-mock/src/core.rs
@@ -18,6 +18,19 @@ use cw_orch_core::{
 
 use super::state::MockState;
 
+pub type MockApp = App<
+    BankKeeper,
+    MockApiBech32,
+    MockStorage,
+    FailingModule<Empty, Empty, Empty>,
+    WasmKeeper<Empty, Empty>,
+    StakeKeeper,
+    DistributionKeeper,
+    IbcSimpleModule,
+    GovFailingModule,
+    StargateFailing,
+>;
+
 /// Wrapper around a cw-multi-test [`App`](cw_multi_test::App) backend.
 ///
 /// Stores a local state with a mapping of contract_id -> code_id/address
@@ -58,22 +71,7 @@ pub struct Mock<S: StateInterface = MockState> {
     /// Inner mutable state storage for contract addresses and code-ids
     pub state: Rc<RefCell<S>>,
     /// Inner mutable cw-multi-test app backend
-    pub app: Rc<
-        RefCell<
-            App<
-                BankKeeper,
-                MockApiBech32,
-                MockStorage,
-                FailingModule<Empty, Empty, Empty>,
-                WasmKeeper<Empty, Empty>,
-                StakeKeeper,
-                DistributionKeeper,
-                IbcSimpleModule,
-                GovFailingModule,
-                StargateFailing,
-            >,
-        >,
-    >,
+    pub app: Rc<RefCell<MockApp>>,
 }
 
 impl<S: StateInterface> Mock<S> {
--- a/packages/cw-orch-mock/src/core.rs
+++ b/packages/cw-orch-mock/src/core.rs
@@ -38,6 +38,11 @@ pub type MockApp = App<
 ///
 /// The state is customizable by implementing the [`StateInterface`] trait on a custom struct and providing it on the custom constructor.
 ///
+/// The addresses used inside this environment are bech32 addresses. For instance, when creating a mock environment
+/// let chain = Mock::new("sender");
+/// the actual sender address can be generated using
+/// let sender_addr = chain.addr_make("sender")
+///
 /// ## Example
 /// ```
 /// # use cosmwasm_std::{Addr, coin, Uint128};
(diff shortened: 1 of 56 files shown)


packages/cw-orch-mock/src/core.rs
```diff
@@ -54,7 +58,22 @@ pub struct Mock<S: StateInterface = MockState> {
     /// Inner mutable state storage for contract addresses and code-ids
     pub state: Rc<RefCell<S>>,
     /// Inner mutable cw-multi-test app backend
-    pub app: Rc<RefCell<BasicApp<Empty, Empty>>>,
+    pub app: Rc<
+        RefCell<
+            App<
```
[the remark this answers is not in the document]
Kayanski: Yes !

packages/cw-orch-mock/src/core.rs
```diff
@@ -145,15 +174,24 @@ impl Mock<MockState> {
...
+    pub fn new_custom(sender: impl Into<String>, prefix: &'static str, custom_state: S) -> Self {
         let state = Rc::new(RefCell::new(custom_state));
-        let app = Rc::new(RefCell::new(custom_app::<Empty, Empty, _>(|_, _, _| {})));
+        let app = Rc::new(RefCell::new(
+            AppBuilder::new_custom()
+                .with_api(MockApiBech32::new(prefix))
+                .with_wasm(WasmKeeper::new().with_address_generator(MockAddressGenerator))
+                .build(|_, _, _| {}),
+        ));
+        let sender = app.borrow().api().addr_make(&sender.into()).clone();
```
Buckram123, line 185: Let's add log of sender address here? @CyberHoward Do you think it would be enough if we cover generated addresses with logs?

CyberHoward: I think a log will be a bit useless as it's logged on creation of the `Mock`. Ideally you want to have the human readable representation of the address whenever the call is made, which won't be easy with this.

I suggest we just leave it as it is and see how annoying it gets.

CyberHoward: However, I do think that this line should be changed. 

If sender is already a valide bech32 we should not re-hash it. 

For example, in our migration tests (clone testing) we set up a mock environment and use the actual wallet address as the sender. With this impl that wallet address will be re-hashed and the resulting address won't exist on-chain.

CyberHoward: So we should use the api's `addr_validate` to check if it's valid and only re-hash the invalid addresses.

packages/cw-orch-mock/src/core.rs
```diff
@@ -370,15 +411,16 @@ impl BankQuerier for Mock {
 impl BankSetter for Mock {
     fn set_balance(
         &mut self,
-        address: &Addr,
+        address: impl Into<String>,
```
CyberHoward, line 414: I used to really like the `impl Into<String>` but now that addresses (`Addr`) must be valid it is too easy to mis-use IMO. 

Opinions?

packages/cw-orch-mock/src/core.rs
```diff
@@ -576,8 +614,8 @@ mod test {
     fn bank_querier_works() -> Result<(), CwEnvError> {
         let denom = "urandom";
         let init_coins = coins(45, denom);
-        let sender = Addr::unchecked("sender");
-        let app = Mock::new(&sender);
+        let app = Mock::new("sender");
+        let sender = app.addr_make("sender");
```
CyberHoward, line 618: Almost every test has these two lines. Maybe `Mock::new` should return `(Mock, Addr)`?

CyberHoward:
A broader question we could ask here is: do we want to enforce use of bech32 or do we want it to be optional? 

If we want it to be optional we should make that clear with the type system. If we want to enforce it we should get rid of most `impl Into<String>` where an address is expected.

packages/cw-orch-mock/src/core.rs
Kayanski: That's a good idea. However, is that predictable for users ? 
For clone testing, we're not using this code anyway, so no issues there :)

packages/cw-orch-mock/src/core.rs
Kayanski: Yeah ... I think that we should avoid using Addr::unchecked() as much as possible now ! Just like in contracts. 
So getting an Addr is mandatory here I think

packages/cw-orch-mock/src/core.rs
CyberHoward: I think I'd be very confused if I gave it a valid bech32 and it turned out to become a different address.

packages/cw-orch-mock/src/core.rs
Kayanski: Don't really like that, when we create a mock, we should get the app only.
If the sender is needed, they can query via `app.sender()` which returns a valid Addr

packages/cw-orch-mock/src/core.rs
CyberHoward: I get why they needed to make this change for instantiate2 but for testing it **sucks**

Kayanski: Yes.
I'm not sure we **have** to make this change though. Investigating

packages/cw-orch-mock/src/core.rs
CyberHoward: actually not really, maybe `Mock::new` can take `&Addr` and a new `Mock::new_str` can take `impl Into<String>` and return a tuple?

PR #313, opened 2024-01-19
Added mock state env id

Kayanski:
This PR aims at adding a `chain-id` parameter to Mock State

Mock State is used with Omsosis test tube and with the `Deploy Details` trait. 

It should therefore be aware of the chain id it supports to be able to deploy the right `DexAdapter` for instance.


CyberHoward:
See https://github.com/AbstractSDK/cw-orchestrator/pull/314#discussion_r1463463888

Kayanski:
> See [#314 (comment)](https://github.com/AbstractSDK/cw-orchestrator/pull/314#discussion_r1463463888)

This is implemented now, what do you think of the names (EnvironmentQuerier and EnvironmentInfo)
Kept the deployment id for state file queries

CyberHoward (approved):

PR #314, opened 2024-01-19
Fix/savings app fixes

Kayanski:
This PR aims at providing necessary tooling for the abstract:savings-app.

Detailed changes : 

1. Added authz querier + fee granter change (https://github.com/AbstractSDK/cw-orchestrator/pull/305)
2. Authz queries implementation + tests (to know whether authz was granted)
3. Change the wasm getter implementation to get the right artifact when deploying the dex adapter. (See this comment : https://github.com/AbstractSDK/cw-orchestrator/commit/555f2799906bb9dc6adcf9dff52d1b249f19ee4e#comments)


packages/cw-orch-core/src/contract/paths.rs
CyberHoward, line 1: How are these changes related to this PR?

packages/cw-orch-mock/src/state.rs
```diff
@@ -21,8 +23,26 @@ impl MockState {
         Self {
             addresses: HashMap::new(),
             code_ids: HashMap::new(),
+            chain_id: mock_env().block.chain_id,
```
CyberHoward, line 26: chain-id is set in `cw_multi_test::App`, how does this relate to that? 

This could (and probably will) cause issues in the future

packages/cw-orch-core/src/contract/paths.rs
Kayanski: This : 
https://github.com/AbstractSDK/cw-orchestrator/commit/555f2799906bb9dc6adcf9dff52d1b249f19ee4e#comments

packages/cw-orch-mock/src/state.rs
Kayanski: Yes, however, the Deploy Details trait necessitates the chain_id field, see below in the same file. It's not a perfect solution, yes !

CyberHoward: So for this we discussed removing this chain_id and updating the `deploy_info` function to use a seperate trait (like `EnvironmentInfo`) that we implement for each env and add as a traitbound to `CwEnv`.

Kayanski:
This was closed by mistake because of a dumb change of branch to merge into. 
The corresponding PR is here : https://github.com/AbstractSDK/cw-orchestrator/pull/322

PR #316, opened 2024-01-23
Nicolas/orc 68 avoid loading node test instances when not doing node tests

Kayanski:
This PR should be merged when tests pass, this is not admissable


Kayanski:
This was pushed directly to main, closing

PR #318, opened 2024-01-23
Generalized querier getter

CyberHoward:
This draft PR shows a different approach to querier construction and access. 

Instead of having a `Getter` trait for each querier we have a single `QuerierGetter` trait that retrieves a querier `Q`. This querier `Q` implements a `Querier` trait. Both are shown below: 

```rust
pub trait QuerierGetter<Q: Querier> {
    fn querier(&self) -> Q;
}

pub trait Querier {
    type Error: Into<CwEnvError> + Debug;
}
```

The inclusion of `type Error` in `Querier` prevents some duplicate code in the traits. 

Now each querier trait needs to add `Querier` as its trait bound, like so: 

```rust
pub trait BankQuerier: Querier {
    fn supply_of(&self, denom: impl Into<String>) -> Result<Coin, Self::Error>;
    // ...
}
```

To get allow for retrieving the querier we implement the `QuerierGetter` with a specified querier like so: 

```rust
impl Querier for DaemonBankQuerier {
    type Error = DaemonError;
}

impl QuerierGetter<DaemonBankQuerier> for Daemon {
    fn querier(&self) -> DaemonBankQuerier {
        DaemonBankQuerier::new(self)
    }
}
```

We can then create arbitrary sets of supported queriers, like: 

```rust
pub trait DefaultQueriers: QuerierGetter<Self::B> + QuerierGetter<Self::W> + QuerierGetter<Self::N> {
    type B: BankQuerier;
    type W: WasmQuerier;
    type N: NodeQuerier;

    fn bank_querier(&self) -> Self::B {
        self.querier()
    }

    fn wasm_querier(&self) -> Self::W {
        self.querier()
    }

    fn node_querier(&self) -> Self::N {
        self.querier()
    }
}
```

And use that as a trait-bound! 

```rust
pub trait QueryHandler: DefaultQueriers {
  // ...
}
```

Thoughts?

### DIFF
--- a/packages/cw-orch-core/src/contract/deploy.rs
+++ b/packages/cw-orch-core/src/contract/deploy.rs
@@ -92,13 +92,12 @@ pub trait Deploy<Chain: CwEnv>: Sized {
         let hash_networks: HashMap<String, (Chain, Self::DeployData)> = networks
             .iter()
             .map(|(c, d)| {
-                Ok::<_, <Chain as QueryHandler>::Error>((
-                    QueryHandler::block_info(c)?.chain_id,
+                Ok::<_, CwEnvError>((
+                    QueryHandler::block_info(c).map_err(Into::into)?.chain_id,
                     (c.clone(), d.clone()),
                 ))
             })
-            .collect::<Result<HashMap<_, _>, _>>()
-            .map_err(Into::into)?;
+            .collect::<Result<HashMap<_, _>, _>>()?;
 
         // First we check the deployment status. Which chains have been un-succesfully deployed since last time
         let chains_to_deploy = if let Ok(deployment_left) = read_deployment() {
--- a/packages/cw-orch-core/src/contract/deploy.rs
+++ b/packages/cw-orch-core/src/contract/deploy.rs
@@ -13,8 +13,8 @@ use std::io::BufReader;
 use std::path::PathBuf;
 
 use crate::env::CwOrchEnvVars;
-use crate::environment::queriers::QueryHandler;
 use crate::environment::CwEnv;
+use crate::environment::QueryHandler;
 use crate::CwEnvError;
 
 use super::interface_traits::ContractInstance;
--- a/packages/cw-orch-core/src/environment/queriers/mod.rs
+++ b/packages/cw-orch-core/src/environment/queriers/mod.rs
@@ -1,7 +1,7 @@
 use cosmwasm_std::{Addr, BlockInfo};
 use serde::{de::DeserializeOwned, Serialize};
 
-use self::{bank::BankQuerierGetter, node::NodeQuerierGetter, wasm::WasmQuerierGetter};
+use self::{bank::{BankQuerier}, node::{NodeQuerier}, wasm::{WasmQuerier}};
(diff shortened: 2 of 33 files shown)


CyberHoward:
@Kayanski do you prefer this over your original design? If so I can proceed with finalizing this.

Kayanski:
> @Kayanski do you prefer this over your original design? If so I can proceed with finalizing this.

yes this is a good idea ! However, I don't see the need for a Querier trait here

packages/cw-orch-core/Cargo.toml
```diff
@@ -25,7 +25,7 @@ eth = ["dep:snailquote"]
 
 [dependencies]
 thiserror = { workspace = true }
-cosmwasm-std = { workspace = true }
+cosmwasm-std.workspace = true
```
Kayanski, line 28: Why this change, is it linting or manual ?

packages/cw-orch-core/src/contract/deploy.rs
```diff
@@ -90,7 +91,12 @@ pub trait Deploy<Chain: CwEnv>: Sized {
     ) -> anyhow::Result<HashMap<String, Self>> {
         let hash_networks: HashMap<String, (Chain, Self::DeployData)> = networks
             .iter()
-            .map(|(c, d)| Ok::<_, Chain::Error>((c.block_info()?.chain_id, (c.clone(), d.clone()))))
+            .map(|(c, d)| {
+                Ok::<_, <Chain as QueryHandler>::Error>((
+                    QueryHandler::block_info(c)?.chain_id,
```
Kayanski, line 96: There is only one block_info function, why indicate that it's the QueryHandler's one ? 

```suggestion
                    c.block_info()?.chain_id,
```

packages/cw-orch-core/src/environment/queriers/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+pub mod node;
+pub mod wasm;
+
+/// This trait acts as the high-level trait bound for supported queries on a `CwEnv` environment.
+/// It also implements some high-level functionality to make it easy to access.
+pub trait QueryHandler: DefaultQueriers {
+    type Error: Into<CwEnvError> + Debug;
+
+    /// Wait for an amount of blocks.
+    fn wait_blocks(&self, amount: u64) -> Result<(), Self::Error>;
```
Kayanski, line 19: I suggest that this function and the others implemented firectly in this trait are simply a wrapper over underlying functions. So we can implemented them inside the trait directly !

CyberHoward: Actually most of them are very env-specific.

CyberHoward: I think block_info is one that can be implemented generic and possibly the wasm query. 

But the Error type is a problem. The trait has its own err type that requires a `Into<CwEnvError>` trait, which the errors returned by the trait (Querier) doesn't have. 

Do we want to add it or?

Kayanski (approved):
LGTM

PR #320, opened 2024-01-26
Fix abstract cw-plus deps

Buckram123:


Cargo.toml
```diff
@@ -14,8 +14,8 @@ cosmwasm-std = { version = "1.1" }
 cw-multi-test = { package = "abstract-cw-multi-test", version = "0.20.2", features = [
   "cosmwasm_1_4",
 ] }
-cw20 = { git = "https://github.com/AbstractSDK/cw-plus.git" }
-cw20-base = { git = "https://github.com/AbstractSDK/cw-plus.git" }
+cw20 = { package = "abstract-cw20", git = "https://github.com/AbstractSDK/cw-plus.git" }
```
CyberHoward, line 17: can't we switch to the actual release of the package instead of the github dep?

Buckram123: Sure, thanks

PR #325, opened 2024-02-08
First impl of Mock and MockBech32

Kayanski:
This PR aims at introducing MockBech32 that enforces valid addresses
Tests needs a little revamp (doc tests mostly), but that's a good looking implementation and usable by users.

### DIFF
--- a/packages/cw-orch-mock/src/core.rs
+++ b/packages/cw-orch-mock/src/core.rs
@@ -2,25 +2,21 @@ use std::{cell::RefCell, fmt::Debug, rc::Rc};
 
 use cosmwasm_std::{
     testing::{MockApi, MockStorage},
-    to_json_binary, Addr, Api, Binary, Coin, CosmosMsg, Empty, Event, Uint128, WasmMsg,
+    to_json_binary, Addr, Api, Binary, CosmosMsg, Empty, Event, WasmMsg,
 };
 use cw_multi_test::{
-    addons::{MockAddressGenerator, MockApiBech32},
-    ibc::IbcSimpleModule,
-    App, AppBuilder, AppResponse, BankKeeper, Contract, DistributionKeeper, Executor,
-    FailingModule, GovFailingModule, StakeKeeper, StargateFailing, WasmKeeper,
+    addons::MockApiBech32, ibc::IbcSimpleModule, App, AppResponse, BankKeeper, Contract,
+    DistributionKeeper, Executor, FailingModule, GovFailingModule, StakeKeeper, StargateFailing,
+    WasmKeeper,
 };
-use cw_utils::NativeBalance;
 use serde::Serialize;
 
 use super::state::MockState;
 use cw_orch_core::{
     contract::interface_traits::Uploadable,
-    environment::{BankQuerier, BankSetter, ChainState, IndexResponse, StateInterface, TxHandler},
+    environment::{ChainState, IndexResponse, StateInterface, TxHandler},
     CwEnvError,
 };
-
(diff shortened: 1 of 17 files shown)


packages/cw-orch-mock/src/core.rs
```diff
@@ -71,113 +70,268 @@ pub type MockApp = App<
 /// // We just use the MockState as an example here, but you can implement your own state struct.
 /// use cw_orch_mock::MockState as CustomState;
 ///
-/// let mock: Mock = Mock::new_custom("sender", CustomState::new());
+/// let mock: Mock = MockBase::new_custom("sender", CustomState::new());
 /// ```
-#[derive(Clone)]
-pub struct Mock<S: StateInterface = MockState> {
+pub struct MockBase<A: Api = MockApi, S: StateInterface = MockState> {
```
CyberHoward, line 75: Do we keep this as pub or make it private?

CyberHoward:
Did a bit of a refactor of the queries to make them API generic by relying on the difference in addr generation between the mock and bech32 impls. 

I'm a bit confused by the `bech32` module in `core.rs`. Can we split these two type impls?

Kayanski:
> Did a bit of a refactor of the queries to make them API generic by relying on the difference in addr generation between the mock and bech32 impls.
> 
> I'm a bit confused by the `bech32` module in `core.rs`. Can we split these two type impls?

Haha gotta love the refactor hack :)

--> Yes we can split the 2 impls in separate files, sure ! The differences are very subtle :)

packages/cw-orch-mock/src/core.rs
Kayanski: We can make it private I think

CyberHoward:
Another question we should ask ourselves is if we want to keep supporting `impl Into<String>` in our address fields?

packages/cw-orch-mock/src/core.rs
Kayanski: Let's leave it private inside the crate for deriving traits.

packages/cw-orch-mock/src/core.rs
```diff
@@ -71,113 +70,268 @@ pub type MockApp = App<
...
+            // We create an address internally
+            let sender = app.borrow().api().addr_make("sender");
+
+            Self { sender, state, app }
+        }
+    }
 
-        chain
+    impl<S: StateInterface> MockBech32<S> {
+        /// Set the bank balance of an address.
```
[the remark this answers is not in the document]
Kayanski: Yes because we are chaging the way the address can be specified. For Bech32, we enforce passing Addr objects while for the normal api, any string will do. 
This simplifies the interface for users not using bech32

packages/cw-orch-mock/src/core.rs
```diff
@@ -71,113 +70,268 @@ pub type MockApp = App<
...
-                for (addr, coins) in balances {
-                    router.bank.init_balance(
-                        storage,
-                        &Addr::unchecked(addr.clone()),
-                        coins.to_vec(),
-                    )?;
-                }
-                Ok(())
-            })
+mod bech32 {
```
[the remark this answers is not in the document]
Kayanski: Yes sorry, some implementation were in the wrong place. It's refactored in different files now

Kayanski:
> Another question we should ask ourselves is if we want to keep supporting `impl Into<String>` in our address fields?

See answer above? I think that it's really good to simplify the users life in case no bech32 verification is needed

CyberHoward (approved):

PR #326, opened 2024-02-12
Fix undefined behavior on simultaneous write in daemon state file

Buckram123:
Simultaneous daemon state file writes currently are not handled. This PR aims to handle such cases.
So there are 2 scenarios when simultaneous write can happen: 
1. Another program that uses daemon state in the process of writing in this file
2. Another daemon of this program tries to write into a file

How this PR tries to fix it:
For the 1.
- We lock a file with `file-lock`, which ensures another program won't be able to write in this file (We don't lock read-only daemonstates)

For the 2.
- We create global state of locked files by this program with `once_cell::sync::Lazy`.
- We disallow other daemons from any thread to access this lock, unless it's cloned directly from the daemon, who originally locked it.
- As soon as every daemon dropped state we unlock file

Beyond safety this PR adds few optimizations:
 - Write json to a file only when all of the state clones done with using this state file
 - Read from a shared state instead of file, when it's in a locked state
 
Closes ORC-75, Closes ORC-84 Closes ORC-140


Kayanski:
This is particularly crucial when using daemons that interact with different chains !

Buckram123:
> This is particularly crucial when using daemons that interact with different chains !

Or with multiple running programs that use daemon. For example constantly running bot and doing deploy of another application

Buckram123:
> It also can corrupt state, in such cases

Update: seems like it was because of mid-way aborted thread.

Buckram123:
@Kayanski wrote a quick proof of concept, that don't break any existing test as well as succeeds the new multi-threaded ones

cw-orch-daemon/src/json_file.rs
```diff
@@ -1,6 +1,86 @@
...
+                "code_ids": {}
+            });
+        }
+    }
+
+    pub fn state(&self) -> Value {
+        self.json.clone()
+    }
+
+    /// Get a value for read
```
CyberHoward, line 61: unused

cw-orch-daemon/src/state.rs
```diff
@@ -148,7 +183,14 @@ impl DaemonState {
     }
     /// Get the state filepath and read it as json
     fn read_state(&self) -> Result<serde_json::Value, DaemonError> {
-        crate::json_file::read(&self.json_file_path)
+        // Check if already open in write mode {
+        let lock = GLOBAL_WRITE_STATE.lock().unwrap();
```
CyberHoward, line 187: so this freezes the thread if the file is open in read mode?

CyberHoward:
I'm a bit confused by the global `Mutex`. It seems like you're using a locking library (`lock-file`) but then added your own locking solution on top. Why did you do that? Could you explain your implementation (preferably in the description of the PR)?

cw-orch-daemon/src/state.rs
Buckram123: > so this freezes the thread if the file is open in read mode?

Yes, it still should be faster if other thread have it open in write mode, and slightly slower otherwise

cw-orch-daemon/src/json_file.rs
Buckram123: > unused

Yes, as stated before, this is proof of concept, and not finished implementation.

Buckram123:
> I'm a bit confused by the global `Mutex`. It seems like you're using a locking library (`lock-file`) but then added your own locking solution on top. Why did you do that? Could you explain your implementation (preferably in the description of the PR)?

Updated PR description! This PR originally had only failing tests with missleading error "EOF while parsing a value", which you can find in `cw-orch-daemon/tests/daemon_state.rs`, to discuss how we want to handle it.

cw-orch-daemon/src/state.rs
Buckram123: added drop of guard in case of reading a file

CyberHoward:
See clippy err

CyberHoward (commented):
So if I understand this PR correctly the JSON state file will be locked when the `Daemon` is constructed? Any other threads that attempt to construct a `Daemon` will be blocked until this original Daemon is dropped?

cw-orch-daemon/src/state.rs
```diff
@@ -9,25 +9,51 @@ use cw_orch_core::{
...
 };
 use ibc_chain_registry::chain::ChainData;
+use once_cell::sync::Lazy;
 use serde::Serialize;
 use serde_json::{json, Value};
-use std::{collections::HashMap, fs::File, path::Path};
+use std::{collections::HashMap, path::Path, sync::Mutex};
 use tonic::transport::Channel;
 
+pub(crate) static GLOBAL_WRITE_STATE: Lazy<Mutex<HashMap<String, (u64, JsonFileState)>>> =
```
CyberHoward, line 18: Could you add a comment here explaining what the types are holding (String, u64)

cw-orch-daemon/src/state.rs
```diff
@@ -9,25 +9,51 @@ use cw_orch_core::{
...
+    json_file_path: String,
     /// Deployment identifier
     pub deployment_id: String,
     /// gRPC channel
     pub grpc_channel: Channel,
     /// Information about the chain
     pub chain_data: ChainData,
     /// Flag to set the daemon state readonly and not pollute the env file
-    pub read_only: bool,
+    read_only: bool,
```
CyberHoward, line 34: When would this flag be used? Could you elaborate on the comment? What is meant with not polluting the env file?

CyberHoward (changes requested):
I see the need for this change but I wonder if this is really the rights solution. 

Imo this change will come back to bite us in the butt at a later point. Yes, it solves the issue of concurrent state writes but it also adds some complexity that could be omitted if we approach the problem from a native async point of view.

The unasked question here is wether we actually want multiple daemons to share the same state? 

If the answer is "in some cases" then we need to outline what those cases are. What's their context? 

I see Daemons sharing state in an async context, most likely a multi-threaded applications where you want to scale request handling and delegate tx execution to different threads. That's an async context where this solution falls short. 

Are there sync contexts that require the same kind of shareable state? I would argue that we shouldn't allow for that because it can introduce weird behavior without the programmer being aware of it.

Let's discuss! @Kayanski @Buckram123

Buckram123:
> So if I understand this PR correctly the JSON state file will be locked when the `Daemon` is constructed? Any other threads that attempt to construct a `Daemon` will be blocked until this original Daemon is dropped?

- File locked after first daemon uses the file
- When using state daemon locks json and unlocks after use
  - Other threads of daemon will be blocked for write/read until previous finished his action
- As soon as all of the daemons dropped "pointer" to state file we write result to file
- If daemon tries to use state file that's already locked by another process it will panic, to prevent hang (see tests)

cw-orch-daemon/src/state.rs
Kayanski: Yes this is used for clone-testing to make sure the dameon is not writing to the `state` file in the clone-testing case, this is a mistake here.

CyberHoward:
> > So if I understand this PR correctly the JSON state file will be locked when the `Daemon` is constructed? Any other threads that attempt to construct a `Daemon` will be blocked until this original Daemon is dropped?
> 
> * File locked after first daemon points to the file
> * When using state daemon locks json and unlocks after use
>   
>   * Other threads of daemons will be blocked for write/read until previous finished his action
> * As soon as all of the daemons dropped "pointer" to state file we write result to file
> * If daemon tries to use state file that's already locked by another process it will panic, to prevent hang (see tests)

Referring to our discussion on Discord. Do we all agree that the second daemon that attempts to access the file should throw an err instead of waiting for the lock to drop?

Buckram123:
> > > So if I understand this PR correctly the JSON state file will be locked when the `Daemon` is constructed? Any other threads that attempt to construct a `Daemon` will be blocked until this original Daemon is dropped?
> > 
> > 
> > 
> > * File locked after first daemon points to the file
> > * When using state daemon locks json and unlocks after use
> >   
> >   * Other threads of daemons will be blocked for write/read until previous finished his action
> > * As soon as all of the daemons dropped "pointer" to state file we write result to file
> > * If daemon tries to use state file that's already locked by another process it will panic, to prevent hang (see tests)
> 
> Referring to our discussion on Discord. Do we all agree that the second daemon that attempts to access the file should throw an err instead of waiting for the lock to drop?

But other Daemons can still use this state file if only it's clone of original daemon, right?

cw-orch-daemon/src/builder.rs
```diff
@@ -48,6 +53,8 @@ impl DaemonAsyncBuilder {
 
     /// Set the deployment id to use for the daemon interactions
     /// Defaults to `default`
+    ///
+    /// This field is ignored for rebuilt daemon and deployment id of the original daemon used instead
     pub fn deployment_id(&mut self, deployment_id: impl Into<String>) -> &mut Self {
         self.deployment_id = Some(deployment_id.into());
         self
```
Buckram123, line 60: I really want to make this editable between "rebuilt" daemons, but can't wrap my head around this

packages/cw-orch-core/src/environment/state.rs
```diff
@@ -14,7 +19,7 @@ pub trait ChainState {
 }
 
 /// This Interface allows for managing the local state of a deployment on any CosmWasm-supported environment.
-pub trait StateInterface: Clone {
+pub trait StateInterface {
```
Buckram123, line 22: We also can't keep `Clone` for StateInterface

CyberHoward: We can if we keep file lock in a global hashmap.

cw-orch-daemon/src/builder.rs
CyberHoward: We shouldn't put the lock in the daemon state. That way we can clone/mutate the key data and not share it between daemon instances.

Buckram123: Kept lock inside DaemonState, but decreased scope to only FullAcess Daemon

packages/cw-orch-core/src/environment/state.rs
Buckram123: left it with hashset, see https://github.com/AbstractSDK/cw-orchestrator/pull/326#discussion_r1595177270

cw-orch-daemon/src/builder.rs
CyberHoward: > Kept lock inside DaemonState, but decreased scope to only FullAcess DaemonState, so it looks like this now:
> 
> ```rust
> enum DaemonStateFile {
>     ReadOnly {
>         path: String,
>     },
>     FullAccess {
>         json_file_state: Arc<Mutex<JsonLockedState>>,
>     },
> }
> ```

Doesn't this allow to read from the file while the state is being written to? I.e. when creating a ReadOnly variant, is there a check that there's no lock on the file? And even if we add that check, we can still create a read-only file and then create a write one.

Buckram123: > > Kept lock inside DaemonState, but decreased scope to only FullAcess DaemonState, so it looks like this now:
> > ```rust
> > enum DaemonStateFile {
> >     ReadOnly {
> >         path: String,
> >     },
> >     FullAccess {
> >         json_file_state: Arc<Mutex<JsonLockedState>>,
> >     },
> > }
> > ```
> 
> Doesn't this allow to read from the file while the state is being written to? I.e. when creating a ReadOnly variant, is there a check that there's no lock on the file? And even if we add that check, we can still create a read-only file and then create a write one.

Lock happens only for writes, we can't do anything if daemon(or something else) mutates file from another program. And dev can manage his read only daemons inside this program (i.e. read before mutations, after mutations or force write to the file to read in the process)

cw-orch-daemon/src/state.rs
```diff
@@ -1,56 +1,69 @@
...
+    pub json_state: DaemonStateFile,
     /// Deployment identifier
     pub deployment_id: String,
-    /// gRPC channel
-    pub grpc_channel: Channel,
-    /// Information about the chain
-    pub chain_data: ChainInfoOwned,
-    /// Flag to set the daemon state readonly and not pollute the env file
-    pub read_only: bool,
+    /// Chain Id
```
CyberHoward, line 30:
```suggestion
    /// Chain Id, ex. osmosis-1
```

cw-orch-daemon/src/state.rs
```diff
@@ -1,56 +1,69 @@
...
+    pub chain_id: String,
+    /// Chain Name
```
CyberHoward, line 32:
```suggestion
    /// Chain Name, ex. "osmosis"
```

CyberHoward (commented):
I'm not entirely sure about moving the channel into the Sender.

IMO it would make more sense to rename Sender -> Signer and then have the Daemon do all the transmitting. That way we can also more easily update the signing logic (like adding multisig or MPC support)

CyberHoward (approved):
LGTM! 
(This is a breaking change)

Buckram123:
@Kayanski @CyberHoward On poisoned Mutex state should we allow saving state in case user called manually `force_write`? Alternatively we can add parameter `ignore_poison` to give more control

Kayanski:
> @Kayanski @CyberHoward On poisoned Mutex state should we allow saving state in case user called manually `force_write`? Alternatively we can add parameter `ignore_poison` to give more control

Let's fail if there's a poison, we can develop the feature if someone asks later

Kayanski (changes requested):
Some changes to clarify things and nitify others

cw-orch-daemon/src/state.rs
```diff
@@ -297,6 +337,7 @@ pub mod test {
             parent_and_relative_state_path
         );
 
+        std::env::remove_var(STATE_FILE_ENV_NAME);
```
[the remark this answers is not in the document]
Kayanski: Maybe yes, if we're keeping state_path

cw-orch-daemon/src/builder.rs
```diff
@@ -87,32 +92,70 @@ impl DaemonAsyncBuilder {
...
     }
 
+    /// Specifies path to the daemon state file
+    /// Defaults to env variable.
+    ///
+    /// Variable: STATE_FILE_ENV_NAME.
+    ///
+    /// This field is ignored for rebuilt daemon and path of the original daemon used instead
+    pub fn state_path(&mut self, path: impl ToString) -> &mut Self {
```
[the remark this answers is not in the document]
Buckram123: For multi-daemon cases it's nice to have this method for flexibility IMO. It gives more straightforward configuration than setting/unsetting env var. 
Where I see it's useful:
- You don't want your daemons fight over file. For example, use separate file for each chain
- Multiple scripts for same contract. For example one script is mainnet and one is testnet and you don't want to use same file for test-chains and main-chains, it also should be fine to run them simultaneously

cw-orch-daemon/src/builder.rs
```diff
@@ -87,32 +92,68 @@ impl DaemonAsyncBuilder {
...
             .chain
             .clone()
             .ok_or(DaemonError::BuilderMissing("chain information".into()))?;
         let deployment_id = self
             .deployment_id
             .clone()
             .unwrap_or(DEFAULT_DEPLOYMENT.to_string());
-        let state = Arc::new(DaemonState::new(chain, deployment_id, false).await?);
+
+        if chain_info.grpc_urls.is_empty() {
```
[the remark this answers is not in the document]
Buckram123: It's not, but I'll move it, sure!

cw-orch-daemon/src/sync/core.rs
```diff
@@ -55,7 +55,7 @@ impl Daemon {
 
     /// Get the channel configured for this Daemon
     pub fn channel(&self) -> Channel {
-        self.daemon.state.grpc_channel.clone()
+        self.daemon.sender.grpc_channel.clone()
```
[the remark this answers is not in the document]
Buckram123: Let's keep it inside sender, until we find sender that don't need grpc channel

cw-orch-daemon/src/state.rs
```diff
@@ -202,7 +241,8 @@ impl StateInterface for DaemonState {
 
     /// Set address for contract in deployment id in state file
     fn set_address(&mut self, contract_id: &str, address: &Addr) {
-        self.set(&self.deployment_id, contract_id, address.as_str())
+        let deployment_id = self.deployment_id.clone();
```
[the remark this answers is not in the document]
Buckram123: self borrowed as mutable

cw-orch-daemon/src/tx_builder.rs
```diff
@@ -25,16 +25,19 @@ pub struct TxBuilder {
...
     // if defined, use this sequence, else get it from the node
     pub(crate) sequence: Option<SequenceNumber>,
+    /// `chain_id` is the unique identifier of the chain this transaction targets.
+    pub(crate) chain_id: String,
 }
 
 impl TxBuilder {
     /// Create a new TxBuilder with a given body.
-    pub fn new(body: Body) -> Self {
+    pub fn new(body: Body, chain_id: String) -> Self {
```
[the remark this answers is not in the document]
Buckram123: Yes, thanks

cw-orch-daemon/src/state.rs
```diff
@@ -133,38 +150,60 @@ impl DaemonState {
...
     /// Retrieve a stateful value using the chainId and networkId
     pub fn get(&self, key: &str) -> Result<Value, DaemonError> {
-        let json = self.read_state()?;
-        Ok(
-            json[&self.chain_data.network_info.chain_name][&self.chain_data.chain_id.to_string()]
-                [key]
-                .clone(),
-        )
+        let json = self.chain_state()?;
+        Ok(json[key].clone())
```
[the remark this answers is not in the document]
Buckram123: Not anymore, moving it back

cw-orch-daemon/src/state.rs
```diff
@@ -133,38 +150,60 @@ impl DaemonState {
...
-    fn read_state(&self) -> Result<serde_json::Value, DaemonError> {
-        crate::json_file::read(&self.json_file_path)
+
+    /// Get the chain state as json
+    fn chain_state(&self) -> Result<serde_json::Value, DaemonError> {
+        let state = match &self.json_state {
+            DaemonStateFile::ReadOnly { path } => {
+                let j = crate::json_lock::read(path)?;
+
+                j[&self.chain_id][&self.chain_name].clone()
```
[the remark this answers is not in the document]
Buckram123: Thanks

cw-orch-daemon/src/builder.rs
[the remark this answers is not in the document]
Kayanski: You're suggesting with this solution for the Daemon to use multple state file for multiple chains, this beats the purpose of the state file which is to have eveyrthing in one place. Its structure is perfect in that way. Especially for multi-chain apps, this is really a must-have.
I don't think it's necessary, because we developped this lock especially to work with the same file !

cw-orch-daemon/src/sync/core.rs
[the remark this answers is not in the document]
Kayanski: agreed

cw-orch-daemon/src/builder.rs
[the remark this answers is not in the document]
Buckram123: Only issue is two different binaries won't be able to access same file because of lock. Anyway it's out of scope of this PR so made it private and unused for now

cw-orch-daemon/src/state.rs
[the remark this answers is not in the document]
Kayanski: ok !

Kayanski (approved):

PR #328, opened 2024-02-14
Use hexbinary as checksum

Buckram123:
This PR aims to return `cosmwasm_std::HexBinary` instead of hex encoded String, for checksums

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ cw-orch-fns-derive = { path = "packages/cw-orch-fns-derive", version = "0.18.0" 
 cw-orch-networks = { path = "packages/cw-orch-networks", version = "0.20.0" }
 
 thiserror = { version = "1.0.21" }
-sha256 = { version = "1" }
+sha2 = { version = "0.10.8" }
 serde_json = "1.0.79"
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -32,7 +32,7 @@ cw-orch-fns-derive = { path = "packages/cw-orch-fns-derive", version = "0.18.0"
 cw-orch-networks = { path = "packages/cw-orch-networks", version = "0.20.0" }
 
 thiserror = { version = "1.0.21" }
-sha256 = { version = "1" }
+sha2 = { version = "0.10.8" }
```
Buckram123, line 35: For using sha2 instead of sha256 - it's lighter(sha2 is dependency of sha256) and returns bytes, instead of String, which is easier to use in pair with HexBinary

CyberHoward (approved):

Buckram123:
@CyberHoward just noticed I forgot to update docs so doc tests are failing 👀

CyberHoward:
> @CyberHoward just noticed I forgot to update docs so doc tests are failing 👀

Can you push a fix to a new pr?

PR #333, opened 2024-02-21
Fix wrong default state file

Buckram123:
Fixes ORC-79. 
Default `state_file` have absolute path, instead of relative. But we only create `~/.cw-orchestrator` dir if path is relative:
https://github.com/AbstractSDK/cw-orchestrator/blob/0f57bf06a740b68d19c60a19ddbce94f2ca857e0/cw-orch-daemon/src/state.rs#L116-L148

This was tested with on https://github.com/AbstractSDK/savings-app/pull/19

### DIFF
--- a/packages/cw-orch-core/src/env.rs
+++ b/packages/cw-orch-core/src/env.rs
@@ -119,7 +119,7 @@ pub fn default_state_folder() -> Result<PathBuf, StdError> {
 impl Default for CwOrchEnvVars {
     fn default() -> Self {
         CwOrchEnvVars {
-            state_file: default_state_folder().unwrap().join("state.json"),
+            state_file: PathBuf::from_str("state.json").unwrap(),
             artifacts_dir: None,
             gas_buffer: None,
             min_gas: None,


packages/cw-orch-core/src/env.rs
```diff
@@ -119,7 +119,7 @@ pub fn default_state_folder() -> Result<PathBuf, StdError> {
 impl Default for CwOrchEnvVars {
     fn default() -> Self {
         CwOrchEnvVars {
-            state_file: default_state_folder().unwrap().join("state.json"),
+            state_file: PathBuf::from_str("state.json").unwrap(),
```
CyberHoward, line 122: I'm a bit confused. Does this make the default behaviour that a relative state file is created, i.e. `my_project/state.json` or `~/.cw-orchestrator/state.json`?

Buckram123: /path/to/home/.cw-orchestrator/ is created when `state_file` path is relative and do not start with `.` or `..`

Buckram123: > `/path/to/home/.cw-orchestrator/` is created when `state_file` path is relative and do not start with `.` or `..`

And it's absolute path without this change

Buckram123: So default of state_file
Before:
`state_file: "/path/to/home/.cw-orchestrator/state.json"`
After:
`state_file: "state.json"`

CyberHoward (approved):

packages/cw-orch-core/src/env.rs
CyberHoward: amazing!

PR #335, opened 2024-02-26
Update test tube to 22.0.0, remove osmosis-std dependency

Kayanski:
This PR aims at updating osmosis_test_tube and rationalizing dependencies.
We've had issues in the past when coupling osmosis test tube and std, let's see if that comes back


CyberHoward:
CI looks beat up on this one

PR #342, opened 2024-03-11
Allow cw-orch wasm compilation without features

Buckram123:
This PR aims to allow adding derive attributes for wasm targets, by not generating any code.

Interface can be removed as well. 
Users can also use the `#[cfg(not(target_arch = "wasm32"))]` to shield the interface instead of the interface feature


contracts/mock_contract/src/custom_resp.rs
```diff
@@ -73,7 +72,3 @@ pub fn migrate(_deps: DepsMut, _env: Env, msg: MigrateMsg) -> StdResult<Response
         ))
     }
 }
-
-#[cfg(feature = "interface")]
-#[cw_orch::interface(InstantiateMsg, ExecuteMsg, QueryMsg, MigrateMsg)]
-struct MockContract;
```
Buckram123, line 79: @Kayanski we can keep it, without any feature/target attributes

Kayanski: Fixed

cw-orch/src/lib.rs
```diff
@@ -6,37 +6,49 @@
...
-#[cfg(feature = "daemon")]
-pub extern crate tokio;
+#[cfg(not(target_arch = "wasm32"))]
 #[cfg(feature = "daemon")]
 pub mod daemon;
 
+#[cfg(not(target_arch = "wasm32"))]
 mod error;
 #[cfg(feature = "osmosis-test-tube")]
 pub mod osmosis_test_tube;
```
CyberHoward, line 20: We should either arch-flag all the exports or only the non-featured ones (if we assume a user doesn't enable the feature in their wasm env). 

Right now daemon is arch-flagged but test-tube is not.

Buckram123: I think arch-flagging features would work as well!

Kayanski: Oups, that's a mistake yes

Kayanski: flaggin features doesn't seem to work : https://github.com/rust-lang/cargo/issues/1197

PR #346, opened 2024-03-14
Removed feature interface from docs

anistark:
This PR aims at updating the cw-orch docs: 
1. Remove the need for the interface feature. Replace by the target flag when needed (only for Uploadable trait)
2. Adapt the docs to remove the mandatory handle parameters inside the Daemon


Kayanski:
We need to keep the re-export though here !

CyberHoward:
I'm pretty sure there are more instances where this flag is used (in the actual code which is referenced in the documentation). 

Maybe something that @Kayanski could remove more efficiently

Buckram123 (approved):
LFGGGGGG

docs/src/contracts/wasm-compilation.md
```diff
@@ -0,0 +1,44 @@
+# Wasm Compilation
+
+`cw-orch` was designed to help you test, deply, script and maintaint your application. None of its features include in-contract operations. cw-orch interfaces, macros and helpers can't be used inside a wasm smart contract.
+
+In case you use one of the cw-orch features inside you smart-contract, the compilation will send an error anyway, so that we are **SURE** you are not using un-intended features inside your contract.
+
+## Importing cw-orch
+
+Importing `cw-orch` as a dependency in your smart contract is not a problem. We have built this library such that every feature that cw-orch imports, uses or exposes is not included when compiled for Wasm. This means that you are safe to use any cw-orch feature when developing your application, creating interfaces, exporting features, because none of it will land in your contract.
```
CyberHoward, line 9:
```suggestion
Importing `cw-orch` as a dependency in your smart contract is not a problem. We have built this library such that every feature that cw-orch imports, uses or exports not included when compiled for Wasm. This means that you are safe to use any cw-orch feature when developing your application, creating interfaces, exporting features, because none of it will land in your contract.
```

docs/src/contracts/wasm-compilation.md
```diff
@@ -0,0 +1,44 @@
...
+```
+
+However, the `Uploadable` traits implementation **IS NOT** safe for smart contracts and needs to import namely `cw-multi-test` elements that we don't remove from WASM compilation. The following code needs to be flagged to not be compiled inside Wasm contracts:
+
+```rust,ignore
+#[cfg(not(target_arch = "wasm32"))]
+{{#include ../../../contracts/counter/src/interface.rs:uploadable_impl}}
+```
+
+### Entry Points
```
CyberHoward, line 34: I think this is the only location where we call them entry points?

CyberHoward (approved):
Some nits on the docs :)

PR #350, opened 2024-03-21
Aether Features

CyberHoward:
Features required for the Abstract GraphQL Rust API. 

- Async query fn generation
- Loosen trait bounds where applicable


cw-orch-daemon/src/core.rs
```diff
@@ -127,6 +127,23 @@ impl DaemonAsync {
         Ok(result)
     }
 
+    /// Query a contract.
+    pub async fn query<Q: Serialize + Debug, T: Serialize + DeserializeOwned>(
```
Kayanski, line 131: this is copy-pasting the `smart_query` from the AsyncWasmQuerier trait. Implement just once please

packages/cw-orch-contract-derive/src/lib.rs
```diff
@@ -250,22 +250,19 @@ pub fn interface(attrs: TokenStream, input: TokenStream) -> TokenStream {
         }
 
         #[cfg(not(target_arch = "wasm32"))]
-        impl<Chain: ::cw_orch::prelude::CwEnv, #all_generics> ::cw_orch::prelude::InstantiableContract for #name<Chain, #all_generics> #all_debug_serialize {
-            type InstantiateMsg = #init;
+        impl<Chain, #all_generics> ::cw_orch::prelude::QueryableContract for #name<Chain, #all_generics> #all_debug_serialize {
```
Kayanski, line 253: Why did you change the order of functions here ?

packages/cw-orch-core/src/environment/queriers/wasm.rs
```diff
@@ -48,3 +48,11 @@ pub trait WasmQuerier: Querier {
         salt: cosmwasm_std::Binary,
     ) -> Result<String, Self::Error>;
 }
+
+pub trait AsyncWasmQuerier: Querier {
+    async fn smart_query<Q: Serialize, T: DeserializeOwned>(
```
Kayanski, line 53: What about those clippy warnings ? Can we use async traits fn in traits now ? Nice

packages/cw-orch-fns-derive/src/query_fns.rs
```diff
@@ -157,5 +157,137 @@ pub fn query_fns_derive(input: ItemEnum) -> TokenStream {
         #derived_trait_impl
     );
 
-    expand.into()
+    expand
+}
+
+pub fn async_query_fns_derive(input: ItemEnum) -> proc_macro2::TokenStream {
```
Kayanski, line 163: Could we have a common function with query_fns_derive, with only some differences (for maintainability) because this seems very close to the sync one

packages/cw-orch-fns-derive/src/query_fns.rs
```diff
@@ -157,5 +157,137 @@ pub fn query_fns_derive(input: ItemEnum) -> TokenStream {
...
+    let derived_trait_impl = quote!(
+        #[automatically_derived]
+        impl<SupportedContract, Chain: ::cw_orch::environment::AsyncWasmQuerier + ::cw_orch::environment::ChainState, #type_generics> #bname<Chain, #type_generics> for SupportedContract
+        #combined_where_clause {}
+    );
+
+    let expand = quote!(
+        #derived_trait
+
+        #derived_trait_impl
```
Kayanski, line 289: For instance here you need to target-flag this derive_trait_impl just like for the sync case. Better merge those into a common function

Kayanski (changes requested):

packages/cw-orch-core/src/environment/queriers/wasm.rs
CyberHoward: https://users.rust-lang.org/t/async-in-public-trait/108400

CyberHoward: Yeah so we should add send bound for sure (as that enables multi-threaded application) 

Also we can use this crate mentioned in that post to create a variant with / without the Send constraint: https://crates.io/crates/trait-variant

We might need this for the async -> sync wrapping that we do (i.e. use the non-send trait in that context). But the compiler will help us figure that out.

Kayanski:
Closing in favor of https://github.com/AbstractSDK/cw-orchestrator/pull/413

PR #354, opened 2024-03-27
Remove the interface features from the docs

Kayanski:
This PR aims at removing the interface feature from the docs. 
Adding some docs along the way if I see the need.


Kayanski:
Close because https://github.com/AbstractSDK/cw-orchestrator/pull/346 already exists

PR #357, opened 2024-04-03
Added Bech32 paragraph

Kayanski:
This PR aims at documenting the MockBech32 structure


Buckram123 (approved):

docs/src/integrations/cw-multi-test.md
```diff
@@ -44,6 +44,30 @@ This environment mocks the actual on-chain execution exactly.
...
+  // Default sender address for this env
+  let sender = mock.sender();
+  ```
+
+- To facilitate the creation of additional valid bech32 addresses, we have added an `addr_make` function. This function uses its string argument to create a new address. That way you can create a new address without having to care about bech32 encoding/decoding yourself:
+
+   ```rust,ignore
+    let named_sender = mock.addr_make("sender");
+    /// This creates a new address with some funds inside of it
+    let rich_sender = mock.addr_make("sender_with_funds", coins(100_000_000_000_000, "ujuno"));
```
CyberHoward, line 68: I think this is wrong? It's the exact same fn as on line 66

PR #361, opened 2024-04-08
Loosen trait bounds in cw-orch-core

Kayanski:
This Pr aims at loosening some trait bounds in cw-orch-core because theyr are un-necessary

### DIFF
--- a/packages/cw-orch-core/src/contract/interface_traits.rs
+++ b/packages/cw-orch-core/src/contract/interface_traits.rs
@@ -1,6 +1,6 @@
 use super::{Contract, WasmPath};
 use crate::{
-    environment::{CwEnv, QueryHandler, TxHandler, TxResponse, WasmQuerier},
+    environment::{ChainState, CwEnv, QueryHandler, TxHandler, TxResponse, WasmQuerier},
     error::CwEnvError,
     log::contract_target,
 };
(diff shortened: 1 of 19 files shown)


packages/cw-orch-core/src/contract/interface_traits.rs
```diff
@@ -206,7 +213,10 @@ pub trait CallAs<Chain: CwEnv>: CwOrchExecute<Chain> + ContractInstance<Chain> +
...
 }
 
-impl<T: CwOrchExecute<Chain> + ContractInstance<Chain> + Clone, Chain: CwEnv> CallAs<Chain> for T {}
+impl<T: CwOrchExecute<Chain> + ContractInstance<Chain> + Clone, Chain: TxHandler> CallAs<Chain>
+    for T
+{
+}
 
 /// Helper methods for conditional uploading of a contract.
 pub trait ConditionalUpload<Chain: CwEnv>: CwOrchUpload<Chain> {
```
CyberHoward, line 222: Can we loosen this bound as well? Could you search for all the uses of `CwEnv` in this library and replace them with their minimal requirements?

Kayanski: That's what I did yes ! And we can't loosen that bound because you need both QueryHandler and TxHandler to be able to use this trait !

Kayanski: emoved in the proc macros as well :)

CyberHoward (approved):
amazing, is this a breaking change to core?

Kayanski:
> amazing, is this a breaking change to core?

I'm not sure how to test that :/
Because we don't have abstract on that version, so it won't be all covering

Kayanski:
> amazing, is this a breaking change to core?

It shouldn't be, but I'm not sure there were not changes in the main branch that were breaking

CyberHoward:
> > amazing, is this a breaking change to core?
> 
> It shouldn't be, but I'm not sure there were not changes in the main branch that were breaking

This is important for you to keep track of. You can go through commits on main and see if any of them are PRs that you expect to be breaking and look into them.

PR #363, opened 2024-04-08
Logs custom CosmTxResponse instead of raw TxResponse

Kayanski:
This PR aims at avoiding to log un-necessary information when debug-logging. 
It just logs the parsed CosmTxResponse instead of the raw TxResponse


CyberHoward:
Did you confirm that this stopped the bin blob to be logged on upload?

Kayanski:
> Did you confirm that this stopped the bin blob to be logged on upload?

Yes !

PR #369, opened 2024-04-15
Added chain names

Kayanski:
This PR aims at adding chain names to cw-orch structures

### DIFF
--- a/packages/cw-orch-networks/src/chain_info.rs
+++ b/packages/cw-orch-networks/src/chain_info.rs
@@ -13,7 +13,11 @@ use cw_orch_core::{
 impl Into<RegistryChainInfo> for ChainInfo<'_> {
     fn into(self) -> RegistryChainInfo {
         RegistryChainInfo {
-            chain_name: self.network_info.id.to_string(),
+            chain_name: match self.kind {
+                ChainKind::Local => self.chain_name.to_string(),
+                ChainKind::Mainnet => self.chain_name.to_string(),
+                ChainKind::Testnet => format!("testnets/{}", self.chain_name),
+            },
             chain_id: self.chain_id.to_string().into(),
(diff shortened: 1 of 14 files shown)


packages/cw-orch-networks/src/chain_info.rs
```diff
@@ -13,7 +13,11 @@ use cw_orch_core::{
 impl Into<RegistryChainInfo> for ChainInfo<'_> {
     fn into(self) -> RegistryChainInfo {
         RegistryChainInfo {
-            chain_name: self.network_info.id.to_string(),
+            chain_name: match self.kind {
+                ChainKind::Local => self.chain_name.to_string(),
+                ChainKind::Mainnet => self.chain_name.to_string(),
+                ChainKind::Testnet => format!("testnets/{}", self.chain_name),
+            },
```
Buckram123, line 20:
```suggestion
            chain_name: self.chain_name,
```
I would prefer to just return it as is, IMO we should have as close result as possible to fetching this `chain_name`. User can add path themself based of `ChainKind`, if he needs to. 
image

Buckram123 (changes requested):
Thanks for adding that, now I can easily resolve testnet explorers for CLI :)
I don't think we should resolve path of testnets for user

packages/cw-orch-networks/src/chain_info.rs
CyberHoward: Agree with this, we shouldn't deflect from chain registry with this.

CyberHoward (changes requested):
I don't think we should do this change. See linear.

Kayanski:
Closed in favor of https://github.com/AbstractSDK/cw-orchestrator/pull/375
See ORC-119

PR #370, opened 2024-04-15
Allow overwriting chain options in Daemon Builder

Kayanski:
This PR aims at allowing developers to overwrite chain options directly in the daemon Builder.

### DIFF
--- a/cw-orch-daemon/src/sync/builder.rs
+++ b/cw-orch-daemon/src/sync/builder.rs
@@ -4,7 +4,7 @@ use crate::{
     DaemonAsyncBuilder,
 };
 use bitcoin::secp256k1::All;
-use ibc_chain_registry::chain::ChainData;
+use ibc_chain_registry::chain::{ChainData, FeeToken, Grpc};
 
 use super::{super::error::DaemonError, core::Daemon};
 
@@ -23,9 +23,12 @@ use super::{super::error::DaemonError, core::Daemon};
 pub struct DaemonBuilder {
     // # Required
     pub(crate) chain: Option<ChainData>,
-    pub(crate) handle: Option<tokio::runtime::Handle>,
     // # Optional
+    pub(crate) handle: Option<tokio::runtime::Handle>,
     pub(crate) deployment_id: Option<String>,
+    pub(crate) additional_grpc_url: Option<String>,
+    pub(crate) gas_denom: Option<String>,
+    pub(crate) gas_fee: Option<f64>,
 
     /* Sender Options */
     /// Wallet sender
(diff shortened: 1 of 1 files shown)


cw-orch-daemon/src/sync/builder.rs
```diff
@@ -23,9 +23,12 @@ use super::{super::error::DaemonError, core::Daemon};
 pub struct DaemonBuilder {
     // # Required
     pub(crate) chain: Option<ChainData>,
-    pub(crate) handle: Option<tokio::runtime::Handle>,
     // # Optional
+    pub(crate) handle: Option<tokio::runtime::Handle>,
     pub(crate) deployment_id: Option<String>,
+    pub(crate) additional_grpc_url: Option<String>,
```
CyberHoward, line 29:
```suggestion
    pub(crate) overwrite_grpc_url: Option<String>,
```

cw-orch-daemon/src/sync/builder.rs
```diff
@@ -97,16 +100,183 @@ impl DaemonBuilder {
         self
     }
 
+    pub fn add_grpc_url(&mut self, url: &str) -> &mut Self {
+        self.additional_grpc_url = Some(url.to_string());
```
CyberHoward, line 104:
```suggestion
        self.overwrite_grpc_url = Some(url.to_string());
```

cw-orch-daemon/src/sync/builder.rs
```diff
@@ -97,16 +100,183 @@ impl DaemonBuilder {
...
 
+        let mut chain = self
+            .chain
+            .clone()
+            .ok_or(DaemonError::BuilderMissing("chain information".into()))?;
+
+        // Override gas fee
+        override_fee(&mut chain, self.gas_denom.clone(), self.gas_fee);
+        // Override grpc_url
+        override_grpc_url(&mut chain, self.additional_grpc_url.clone());
```
CyberHoward, line 129:
```suggestion
        override_grpc_url(&mut chain, self.overwrite_grpc_url.clone());
```

cw-orch-daemon/src/sync/builder.rs
CyberHoward: Should either rename this or update the URL logic to *add* the url to the grpc url list instead of replacing it alltogether.

Kayanski: We need to be able to replace it in case the current one gives errors (even if we have fallback logic right now)
Choosing for replacing it (works in most cases)

CyberHoward:
stable clippy is failing

CyberHoward (approved):

PR #373, opened 2024-04-17
Separating env variables in respective crates

Kayanski:
This PR aims at separating env variables in their respective crates.
Closes ORC-120


Buckram123 (approved):
Separating them is nice

cw-orch-daemon/src/env.rs
```diff
@@ -0,0 +1,124 @@
...
+            disable_wallet_balance_assertion: false,
+            disable_logs_message: false,
+        }
+    }
+}
+
+impl DaemonEnvVars {
+    pub fn load() -> Result<Self, CwEnvError> {
+        let mut env_values = DaemonEnvVars::default();
+
```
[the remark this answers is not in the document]
CyberHoward: Agreed, would be nice to use cell here.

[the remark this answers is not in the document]
Kayanski: Using OnceCell or lazy_static here means it's not possible to change those env variables in code, once it's initialized. 
Because this is hidden from the user, unexpected behavior can happen which would be hard to debug for the user. 
I would prefer the second solution !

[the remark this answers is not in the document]
CyberHoward: I wouldn't lazy static it but initialize it with the daemon state instead. That way the user can still execute code (and set envs) before this is loaded.

cw-orch-daemon/src/env.rs
```diff
@@ -0,0 +1,133 @@
...
+    env_var_name: &str,
+) -> F {
+    value
+        .parse()
+        .map_err(|e| {
+            StdError::generic_err(format!(
+                "Couldn't parse content of env var {env_var_name}, error : {e}"
+            ))
+        })
+        .unwrap()
```
Buckram123, line 132: It's not really `cosmwasm_std::StdError` and we panic on error anyway, let's use `.expect()` or just `panic!()` here instead?

cw-orch-daemon/src/lib.rs
```diff
@@ -23,6 +24,7 @@ pub mod tx_builder;
 pub use self::{builder::*, channel::*, core::*, error::*, state::*, sync::*, tx_resp::*};
 pub use cw_orch_networks::chain_info::*;
 pub use cw_orch_networks::networks;
+pub use env::DaemonEnvVars;
```
Buckram123, line 27: Do we need to re-export it if it's already exposed under `env`?

Buckram123 (approved):
Nice! :rocket:

cw-orch-daemon/src/lib.rs
CyberHoward: Yeah, it's mostly used internally so shouldn't need re-export.

cw-orch-daemon/src/env.rs
```diff
@@ -0,0 +1,133 @@
...
+use cosmwasm_std::StdError;
+
+const DEFAULT_TX_QUERY_RETRIES: usize = 50;
+
+pub const STATE_FILE_ENV_NAME: &str = "STATE_FILE";
+pub const GAS_BUFFER_ENV_NAME: &str = "CW_ORCH_GAS_BUFFER";
+pub const MIN_GAS_ENV_NAME: &str = "CW_ORCH_MIN_GAS";
+pub const MAX_TX_QUERIES_RETRY_ENV_NAME: &str = "CW_ORCH_MAX_TX_QUERY_RETRIES";
+pub const MIN_BLOCK_SPEED_ENV_NAME: &str = "CW_ORCH_MIN_BLOCK_SPEED";
+pub const DISABLE_WALLET_BALANCE_ASSERTION_ENV_NAME: &str =
```
CyberHoward, line 20: Maybe a nit but are we defaulting to having the env-var named as `disable` and basically defaulting to setting env-vars to true if they want to enable something (so default all our bool env vars to false) or do we change it to the env-vars always indicating the feature and having the env var toggle them. I.e. rename these "disable" env vars to "wallet_balance_assertion", "log_activation", etc. and having them set to a default value.

PR #381, opened 2024-04-25
Using logs and events in Tx Response

Kayanski:
This PR aims at fixing an update pushed by several changes to not include events on the TxResponse logs field.

When events fetched from logs are empty, the function will also look at events to find corresponding events


cw-orch-daemon/examples/event-query.rs
```diff
@@ -0,0 +1,24 @@
+use cw_orch_core::environment::DefaultQueriers;
+use cw_orch_core::environment::NodeQuerier;
+use cw_orch_daemon::DaemonBuilder;
+use cw_orch_networks::networks;
+
+const LOCAL_MNEMONIC: &str = "clip hire initial neck maid actor venue client foam budget lock catalog sweet steak waste crater broccoli pipe steak sister coyote moment obvious choose";
```
Buckram123, line 6:
```suggestion
// https://github.com/CosmosContracts/juno/blob/32568dba828ff7783aea8cb5bb4b8b5832888255/docker/test-user.env#L2
const LOCAL_MNEMONIC: &str = "clip hire initial neck maid actor venue client foam budget lock catalog sweet steak waste crater broccoli pipe steak sister coyote moment obvious choose";
```
Would be nice to add reference or comment on random mnemonics, so reviewer(or anyone who steps on it next time in the code) won't check if it's public or you accidentally leaked something haha

Kayanski: Hmm, good idea !

cw-orch-daemon/src/tx_resp.rs
```diff
@@ -90,6 +90,33 @@ impl CosmTxResponse {
...
+                })
+                .collect::<Vec<_>>();
+
+            events_filtered
+        } else {
+            log_events
+        }
+    }
+
+    pub fn get_events_from_logs(&self, event_type: &str) -> Vec<TxResultBlockEvent> {
```
CyberHoward, line 119: Do these both need to be pub?

PR #385, opened 2024-05-07
Moved osmosis test tube

Kayanski:
This Pr aims at moving osmosis test tube outside of cw-orch.
This is now a package that is independent and depends on cw-orch


packages/cw-orch-osmosis-test-tube/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[dependencies]
+
+# Test Tube env deps
+osmosis-test-tube = { workspace = true }
+prost-types = { workspace = true }
+prost = { workspace = true }
+cosmwasm-std.workspace = true
+serde.workspace = true
+cw-orch = { workspace = true }
```
CyberHoward, line 19: Can we make this so it only depends on `cw-orch-core`. Ideally cw-orch is just a crate that ships with the default environments that you need (daemon and mock) and other environments can be plugged into it (without those envs having to depend on cw-orch itself). That way they can be independently versioned.

CyberHoward (changes requested):
Would love to see if / how we can keep the test-tube crate independent from `cw-orch` crate.

packages/cw-orch-osmosis-test-tube/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+[dependencies]
+
+# Test Tube env deps
+osmosis-test-tube = { workspace = true }
+prost-types = { workspace = true }
+prost = { workspace = true }
+cosmwasm-std.workspace = true
+serde.workspace = true
+cw-orch-core = { workspace = true }
+cw-orch-mock = { workspace = true }
```
CyberHoward, line 20: It needs mock for the state struct?

CyberHoward: Maybe state should be its own package so it can be versioned / developed independently? Might not be worth the hassle

PR #387, opened 2024-05-13
Daemon Sender separation

Kayanski:
This PR aims at separating the daemon and the sender to be able to compose with senders and queriers.

I removed the Arc wrapper around the Sender because the `SenderTrait` trait needs the `set_options` method to take a mutable reference, which arc can't do because it can't mutate the underlying object. 

When using a Mutex to allow for mutability, the async methods of the trait error because of a Send bound that is not available.

We will need to think about that, if we want the Sender to be copiable


cw-orch-daemon/src/senders/multiple_sender.rs
```diff
@@ -0,0 +1,116 @@
...
+use cosmrs::{AccountId, Any};
+use cosmwasm_std::Addr;
+use cw_orch_core::log::transaction_target;
+use prost::Name;
+
+use std::sync::{Arc, Mutex};
+
+use super::{base_sender::SenderOptions, sender_trait::SenderTrait};
+
+pub type MultiDaemon = DaemonBase<MultipleSender>;
```
CyberHoward, line 15: Can we rename `MultiDaemon` to `BatchDaemon` or `MultiTxDaemon` 

Same with `MultiSender` -> `BatchSender` / `MultiTxSender`?

cw-orch-daemon/tests/authz.rs
```diff
@@ -63,6 +63,8 @@ mod tests {
             authorization: Some(authorization.clone()),
             expiration: Some(expiration.clone()),
         };
+
+        println!("{:?}-{:?}", sender, grantee);
```
CyberHoward, line 67: rm

cw-orch-daemon/src/senders/manual_sender.rs
```diff
@@ -0,0 +1,178 @@
...
+    ) -> Result<CosmTxResponse, DaemonError> {
+        // We simulate
+        let gas_needed = self.simulate(msgs, memo).await?;
+
+        log::info!("Here is the transaction to sign and broadcast: ");
+        log::info!("Gas needed: {}", gas_needed);
+
+        log::info!("Enter the txhash to proceed");
+        let stdin = io::stdin();
+        let txhash = stdin.lock().lines().next().unwrap().unwrap();
```
Buckram123, line 56: Nit: you can use read_line for getting 1 line https://doc.rust-lang.org/std/io/trait.BufRead.html#method.read_line

cw-orch-daemon/src/senders/manual_sender.rs
```diff
@@ -0,0 +1,178 @@
...
+
+#[derive(Clone, Default)]
+pub struct ManualSenderOptions {
+    pub sender_address: Option<String>,
+}
+
+/// Signer of the transactions and helper for address derivation
+/// This is the main interface for simulating and signing transactions
+#[derive(Clone)]
+pub struct ManualSender {
```
Buckram123, line 33: What's the purpose of manual sender?

cw-orch-daemon/src/senders/multiple_sender.rs
```diff
@@ -0,0 +1,116 @@
...
+
+/// Signer of the transactions and helper for address derivation
+/// This is the main interface for simulating and signing transactions
+#[derive(Clone)]
+pub struct MultipleSender {
+    /// Contains the different messages to broadcast
+    /// These are behind an Arc Mutex, because `commit_tx_any function` doesn't have access to a mutable reference to the object
+    pub msgs: Arc<Mutex<Vec<Any>>>,
+    pub sender: Wallet,
+}
```
Buckram123, line 25: Nice, would love to use it for batch actions in CLI :rocket:

cw-orch-daemon/src/senders/sender_trait.rs
```diff
@@ -0,0 +1,52 @@
...
+use tonic::transport::Channel;
+
+use crate::{CosmTxResponse, DaemonError};
+
+pub trait SenderTrait: Clone {
+    type Error: Into<DaemonError> + std::error::Error + std::fmt::Debug + Send + Sync + 'static;
+    type SenderOptions: Default + Clone;
+
+    // TODO: do we want to enforce sync on this function ?
+    fn address(&self) -> Result<Addr, Self::Error>;
```
Buckram123, line 13: Is there a case where we would prefer async? Multisigner?

Buckram123 (commented):
Looks really nice, would like to know more about why we need `ManualSender`!

cw-orch-daemon/src/builder.rs
```diff
@@ -36,60 +31,89 @@ pub struct DaemonAsyncBuilder {
...
+    ) -> DaemonAsyncBuilderBase<OtherSenderGen> {
+        DaemonAsyncBuilderBase {
+            chain: self.chain.clone(),
+            deployment_id: self.deployment_id.clone(),
+            state_path: self.state_path.clone(),
+            state: self.state.clone(),
+            sender: Some(wallet),
+            sender_options: OtherSenderGen::SenderOptions::default(),
+            write_on_change: self.write_on_change,
+        }
```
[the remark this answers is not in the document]
Kayanski: I can't mutate, because it's another object so I can't change its type. I can't take `self` either because it will throw a temporary object error. That's the only solution that we have with our `&mut self`-> `&mut Self` builder pattern

cw-orch-daemon/src/core.rs
```diff
@@ -317,8 +337,14 @@ impl DaemonAsync {
...
-    pub fn set_sender(&mut self, sender: &Wallet) {
-        self.sender = sender.clone();
+    pub fn set_sender<SenderGen2: SenderTrait>(
+        self,
+        sender: SenderGen2,
+    ) -> DaemonAsyncBase<SenderGen2> {
+        DaemonAsyncBase {
+            sender,
+            state: self.state,
+        }
```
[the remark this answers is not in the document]
Kayanski: It's generic to not conflict with the `Sender` structure

cw-orch-daemon/src/core.rs
```diff
@@ -30,6 +30,10 @@ use std::{
 
 use tonic::transport::Channel;
 
+use crate::senders::sender_trait::SenderTrait;
+
+pub const INSTANTIATE_2_TYPE_URL: &str = "/cosmwasm.wasm.v1.MsgInstantiateContract2";
+
```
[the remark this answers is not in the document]
Kayanski: The update is more complicated than that, it should stays a task

cw-orch-daemon/src/core.rs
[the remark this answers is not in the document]
Kayanski: but changed

cw-orch-daemon/src/senders/sender_trait.rs
Kayanski: I don't think so for now. Let's think about this later when we need it ? Removing the todo

cw-orch-daemon/src/senders/manual_sender.rs
Kayanski: It's there to be able to broadcast a transaction using a multisig for instance, or when you want to send a transaction using other means. This is meant as an example to showcase the use of a Sender in this way, do you think it's worth keeping ?

CyberHoward: If it's an example, could you rename the file to `example_sender.rs` or similar?

CyberHoward: Or include it in `examples` folder instead of the bin?

Kayanski: Done now

cw-orch-daemon/src/builder.rs
```diff
@@ -167,40 +191,32 @@ impl DaemonAsyncBuilder {
...
-                }
-            },
-            None => Sender::new_with_options(
-                chain_info.clone(),
-                GrpcChannel::connect(&chain_info.grpc_urls, &chain_info.chain_id).await?,
-                sender_options,
-            )?,
+        let sender = if let Some(sender) = &self.sender {
+            let mut sender = sender.clone();
+            sender.set_options(self.sender_options.clone());
```
CyberHoward, line 197: I don't think this will give use the expected behaviour. 

Let's say we want to use a custom sender. First, we're required to build the sender ourselves (there's no way to set the custom sender options ATM). So if we build it and configure it to our liking, then we'll provide it to this builder. The builder will take and and then **apply the `SenderOptions::Default` value to the sender that we just configured. This would be extremely confusing to run into. 

Additionally if we add the ability to set the options for the sender, we can still confuse the user with this behaviour. 

And it its current state the `else` part of this builder will only be called with `Sender` being the default `Wallet`. The `Sender::build` function is only called in this location and adds a lot of complexity to the `Sender` trait.

CyberHoward:
Closed in favor of #438

PR #388, opened 2024-05-13
Added backwards compatibility-test

Kayanski:
This PR aims at adding a contract to test cw-orch backwards compatibility on the provided macros.
This is very useful to make sure that the cw-orch-core stabilitization was useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- Added a test to make sure the derive macros stay compatible with new cw-orch versions
+
+## 0.22
+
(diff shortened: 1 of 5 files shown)


CyberHoward (commented):
Could we try and do a `0.23.0-alpha` release just to see if this works?

CyberHoward:
Also we should be able to do a lot more frequent releases now :)))

Kayanski:
> Could we try and do a `0.23.0-alpha` release just to see if this works?

Did that with a aaa-test--cw-orch crate instead haha, that was a better idea !

PR #392, opened 2024-05-14
Use only cw-orch-core inside derive macros

Kayanski:
This PR aims at removing the dependency on cw-orch structure for the derives macros.
They now only depend on cw_orch::core. This re-export of cw-orch-core should NEVER be touched


CyberHoward (approved):
nice!

PR #389, opened 2024-05-13
Added automatic into for msgs

Kayanski:
This PR aims at adding automatic into traits for macros

Closes ORC-132

PR #398, opened 2024-05-17
Merge Execute and Query derive Macros

Kayanski:
This PR aims at removing duplicate code and merges execute and query msgs.

# Tasks done

- Merges query and Execute derive functions
- Rationalized the macros in order to fully use syn instead of home-made functions and manipulations 
   --> This was done to be able to better support generics in the future.

Depends on https://github.com/AbstractSDK/cw-orchestrator/pull/389


Buckram123 (approved):
LGTM

PR #399, opened 2024-05-17
Simplify the derived function arguments

Kayanski:
This Simplify the cw-orch contract function derives.

The functions do not take their argument types directly but an argument of the form `Into<type>` This allows for simpler interactions

Depends on https://github.com/AbstractSDK/cw-orchestrator/pull/398

## Downside
The types have to be known when specifying the argument. 
This is specifically bad when using `collect`. They should be changed to `collect::Vec<_>()`  for instance

## Breaking changes
Here are the cases that WILL break when releasing this PR.
Cw-orch-core is not breaking

cw-orch-fns-derive breaks

```rust
 // The received type is impl Into<Uint128>, so the type is impossible to derive
contract.mint(1278u128.into());
// Replaced by 
contract.mint(1278u128);
```

```rust
 // The received type is impl Into<Uint128>, so the type is impossible to derive
let mut receivers = HashSet::new();
receivers.insert("kayanski");
receivers.insert("misha");
// Just a conversion from HashSet to Vec
contract.multi_send(receivers.into_iter().collect());
// Replaced by 
contract.mint(receivers.into_iter().collect::<Vec<_>>());
```

Depends on https://github.com/AbstractSDK/cw-orchestrator/pull/398


contracts/mock_contract/test.rs
Buckram123, line 1: Do I need to read this file?

cw-orch/tests/impl_into.rs
```diff
@@ -46,7 +46,7 @@ impl<Chain: CwEnv> Example<Chain> {
         self.bar("hello".to_string()).unwrap();
 
         // function `test` is available because of the `impl_into` attribute!
```
Buckram123, line 48: Update comments

Buckram123 (approved):
Looks nice, shouldn't it require minor bump instead of patch of `cw-orch-fns-derive`? I guess missing
- [ ] Changelog
- [ ] Docs update where impl_into mentioned

Not sure about the file that I skipped, looked like `cargo expand` output, lmk if I should check it as well

CyberHoward (commented):
I don't think we should add `impl Into` for every type of argument. 

I suggest we add a match statement that only converts the arguments to `impl Into<>` for the types: 

- Uint (all sizes)
- String

But not for 
- Addr
- Custom structures
- booleans

This helps us keep the generated fns readable (as you can only see the generated code in your IDE).

Kayanski:
> I don't think we should add `impl Into` for every type of argument.
> 
> I suggest we add a match statement that only converts the arguments to `impl Into<>` for the types:
> 
>     * Uint (all sizes)
> 
>     * String
> 
> 
> But not for
> 
>     * Addr
> 
>     * Custom structures
> 
>     * booleans
> 
> 
> This helps us keep the generated fns readable (as you can only see the generated code in your IDE).

Yes good idea ! 
We just need to keep in mind that because of how proc-macros work, we will not be able to accept such bounds in types like : 
```
pub type ChainId = String;
```

Buckram123:
> > I don't think we should add `impl Into` for every type of argument.
> > I suggest we add a match statement that only converts the arguments to `impl Into<>` for the types:
> > ```
> > * Uint (all sizes)
> > 
> > * String
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But not for
> > ```
> > * Addr
> > 
> > * Custom structures
> > 
> > * booleans
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This helps us keep the generated fns readable (as you can only see the generated code in your IDE).
> 
> Yes good idea ! We just need to keep in mind that because of how proc-macros work, we will not be able to accept such bounds on type aliases like :
> 
> ```
> pub type ChainId = String;
> ```

Is it possible to check it with typeid? https://doc.rust-lang.org/std/any/struct.TypeId.html

Kayanski:
> Is it possible to check it with typeid? https://doc.rust-lang.org/std/any/struct.TypeId.html

Undortunately not :/
See : https://users.rust-lang.org/t/get-type-information-in-proc-macro/48909

CyberHoward:
> > I don't think we should add `impl Into` for every type of argument.
> > I suggest we add a match statement that only converts the arguments to `impl Into<>` for the types:
> > ```
> > * Uint (all sizes)
> > 
> > * String
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But not for
> > ```
> > * Addr
> > 
> > * Custom structures
> > 
> > * booleans
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This helps us keep the generated fns readable (as you can only see the generated code in your IDE).
> 
> Yes good idea ! We just need to keep in mind that because of how proc-macros work, we will not be able to accept such bounds on type aliases like :
> 
> ```
> pub type ChainId = String;
> ```

Think that's fine. We could add a field attribute (like `payable`) that enables the Into manually as well. 

Maybe we should start prefixing our macro attributes as well, like serde? 

```rust
#[cw_orch(payable)]

#[cw_orch(into)]
```

Buckram123:
> Maybe we should start prefixing our macro attributes as well, like serde?
> 
> ```rust
> #[cw_orch(payable)]
> 
> #[cw_orch(into)]
> ```

Pretty sure we should if we use such common phrases as `into` haha

Kayanski:
> > Maybe we should start prefixing our macro attributes as well, like serde?
> > ```rust
> > #[cw_orch(payable)]
> > 
> > #[cw_orch(into)]
> > ```
> 
> Pretty sure we should if we use such common phrases as `into` haha

Yes there is a PR with that here already : https://github.com/AbstractSDK/cw-orchestrator/pull/408

PR #401, opened 2024-05-21
Add flush state function for local daemons

Kayanski:
This PR aims at introducing a state flushing method for local chain Daemons


cw-orch-daemon/src/state.rs
```diff
@@ -167,6 +167,25 @@ impl DaemonState {
...
+        if self.chain_data.kind != ChainKind::Local {
+            panic!("Can only flush local chain state");
+        }
+        if self.read_only {
+            return Err(DaemonError::StateReadOnly);
+        }
+
+        let mut json = self.read_state()?;
+
+        json[&self.chain_data.network_info.chain_name][&self.chain_data.chain_id.to_string()] =
```
Buckram123, line 183: I can see it be running at the start of "local" scripts. I think this will panic if there are no local state yet. I think we should try to do multiple `.get_mut()` and do nothing if there are no state

Buckram123 (changes requested):
Cool feature! Need to make it work without state

cw-orch-daemon/src/state.rs
Kayanski: This doesn't panic (it just returns Value::Null) :)

Kayanski: But let me make sure

Kayanski: This doesn't panic, added to the examples if you want to run it

Buckram123: > This doesn't panic (it just returns Value::Null) :)

Oh, that's nice, didn't know that, sorry

Buckram123 (approved):
LGTM

PR #402, opened 2024-05-22
Helpers to querying raw state outside contracts

Kayanski:
This PR aims at adding a simple helper to query raw state from outside contracts.
Types are derived automatically.

You can use the following syntax : 
```rust
const TEST_ITEM: Item<TestItem> = Item::new("test-item");
const TEST_MAP: Map<String, TestItem> = Map::new("test-map");

let contract = Contract::new(chain);
contract.item_query(TEST_ITEM);
contract.map_query(MAP_ITEM, map_key);
```

Do you think we should use this instead  : 
```rust
TEST_ITEM.cw_orch_query(&contract)?;
MAP_ITEM.cw_orch_query(&contract, map_key)?;
```

### DIFF
--- a/packages/cw-orch-core/src/environment/queriers/wasm.rs
+++ b/packages/cw-orch-core/src/environment/queriers/wasm.rs
@@ -1,5 +1,5 @@
 use cosmwasm_std::{from_json, CodeInfoResponse, ContractInfoResponse, HexBinary};
-use cw_storage_plus::Item;
+use cw_storage_plus::{Item, Map, PrimaryKey};
 use serde::{de::DeserializeOwned, Serialize};
 
 use crate::{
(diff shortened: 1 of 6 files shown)


packages/cw-orch-core/src/environment/queriers/wasm.rs
```diff
@@ -27,6 +28,30 @@ pub trait WasmQuerier: Querier {
...
+        &self,
+        address: impl Into<String>,
+        item: Item<T>,
+    ) -> Result<T, CwEnvError> {
+        let current_manager_version = self
+            .raw_query(address, item.as_slice().to_vec())
+            .map_err(Into::into)?;
+
+        from_json(current_manager_version).map_err(Into::into)
+    }
```
Buckram123, line 41: Should we add to it a method that can query any key (Vec<u8>) to support any types of state objects, including custom?

Buckram123 (approved):
Nice syntax! Item and Map are the most common state objects, but it could be nice to have queries of any raw key as well. Easy to implement as well

packages/cw-orch-core/src/environment/queriers/wasm.rs
Buckram123: > What do you mean ? This is the raw query function here, it's already available. I have to rename those objects though, wtf are those names haha

I mean on contract object: `item_query(ITEM)`, `map_query(MAP, KEY)` and `raw_state_query(KEY)`(not sure about the name)

Kayanski: Ok, done

PR #413, opened 2024-06-09
Added async querier fns derive

Kayanski:
This PR aims at adding async query compatibility to contracts to be able to query them asynchronously.

This paves the way for transaction execution asyncronous. 
However, for Execution, the Sender has to be mutex-ed to avoid sequence conflicts

### Checklist

- [x] Changelog updated.
- [x] Docs updated.


CyberHoward (approved):

Buckram123 (approved):
Very cool!

PR #423, opened 2024-06-17
Nicolas/orc 144 rename the get chain helper

Kayanski:
This PR aims at adding an `environment` method on contracts that replaces the current `get_chain`.
We also add an `Environment` trait that allows any structure to be built around an `environment`

This is not breaking ? Should we make the change at the Cw-Orch-Core level and make that breaking ? 

### Checklist

- [ ] Changelog updated.
- [ ] Docs updated.


Buckram123:
> This is not breaking ? Should we make the change at the Cw-Orch-Core level and make that breaking ?

We can deprecate `.get_chain()` method for the patch and remove from cw-orch-core in the next version. What do you think?

Kayanski:
> > This is not breaking ? Should we make the change at the Cw-Orch-Core level and make that breaking ?
> 
> We can deprecate `.get_chain()` method for the patch and remove from cw-orch-core in the next version. What do you think?

Yes good idea, what about the `Environment` trait ? We can't use it and make this change non-breaking

packages/cw-orch-core/src/environment/mod.rs
```diff
@@ -17,3 +17,8 @@ pub use queriers::{
     DefaultQueriers, Querier, QuerierGetter, QueryHandler,
 };
 pub use state::{ChainState, StateInterface};
+
```
CyberHoward, line 20: Update doc-comment(s)

packages/cw-orch-core/src/contract/interface_traits.rs
```diff
@@ -188,12 +192,19 @@ pub trait CwOrchQuery<Chain: QueryHandler + ChainState>:
...
+        self.environment()
             .wasm_querier()
             .map_query(self.address()?, query_map, key)
     }
 }
 
+impl<Chain: ChainState, T: ?Sized + ContractInstance<Chain>> Environment<Chain> for T {
+    fn environment(&self) -> &Chain {
+        #[allow(deprecated)]
+        self.get_chain()
```
CyberHoward, line 204: kind of weird to use a deprecated function here. 

If we sunset this fn in the trait we should also sunset it in the `Contract` type.

So update the old fn: 
```rust
fn get_chain(&self) -> &Chain {
        self.as_instance().environment()
}
```

And then have the same code here: 

```rust
fn get_chain(&self) -> &Chain {
        self.as_instance().environment()
}
```

And implement Environment for Contract.

Kayanski: I can't implement Environment for Contract, because of this compiler error : 
 `downstream crates may implement trait 'contract::interface_traits::ContractInstance<_>' for type 'contract::contract_instance::Contract<_>'`

So I decided to just have a function in the impl rather than implementing the trait

Buckram123 (approved):
LGTM

PR #424, opened 2024-06-18
Fix state reuse on a new chains

Buckram123:
Just noticed that there is a bug regarding reused state. State won't be prepared for the writes if new chain is different from the previous one, unless it was prepared before.
This PR adds `state.prepare()`, if Daemon Builder uses existing state 
### Checklist

- [ ] Changelog updated.
- [ ] Docs updated.


Kayanski (approved):
LGTM

PR #438, opened 2024-07-01
Refactor the `Sender` creation and `Sender` traits

CyberHoward:
This PR refactors the way that the Sender objects are constructed and configured. 

The initial PR added a `SenderTrait::SenderOptions` associated type that allow for building a `Sender` object. 

First, I renamed a few of the traits, types and methods to improve the clarity of the code's structure. Then I split up the `SenderTrait` into `SenderBuilder`, `QuerySender` and `TxSender`, each with their own scoped functions and types. 

With this I refactored the `DaemonBuilder`(s) to now work as follows: 

```rust
// Creates a Daemon with the env-var mnemonic. Minimal code required
let daemon = Daemon::builder(chain).mnemonic("...").build()?; // --> Chain now required to build `Builder`

// Creates a `Wallet` with a mnemonic
// Can also customize the `CosmosOptions` to use authz, different hd path, etc.
let my_mnemonic = "...";
let daemon = Daemon::builder(chain).build_sender(CosmosOptions::new(my_mnemonic))?;

// Or create a custom `Sender` or query-only sender
let daemon: QueryOnlyDaemon = Daemon::builder(chain).build_sender(())?;
```

### Checklist

- [x] Changelog updated.
- [ ] Docs updated.


cw-orch-daemon/src/builder.rs
```diff
@@ -21,76 +24,30 @@ pub const DEFAULT_DEPLOYMENT: &str = "default";
...
-        self
-    }
-}
-
-impl<Sender: SenderTrait> DaemonAsyncBuilderBase<Sender> {
-    /// Set the chain the daemon will connect to
-    pub fn chain(&mut self, chain: impl Into<ChainInfoOwned>) -> &mut Self {
-        self.chain = Some(chain.into());
-        self
-    }
```
Buckram123, line 93: Removing `.chain()` from the builder makes it harder to use same daemon or daemonbuilder for different chains daemons.

Maybe just adding `chain: Option<ChainInfoOwned>` to rebuild would be enough

cw-orch-daemon/src/channel.rs
```diff
@@ -90,6 +90,11 @@ impl GrpcChannel {
 
         Ok(successful_connections.pop().unwrap())
     }
+
+    /// Create a gRPC channel from the chain info
+    pub async fn from_chain_info(chain_info: &ChainInfoOwned) -> Result<Channel, DaemonError> {
```
Buckram123, line 95: small nit: IMO should be possible to use either ChainInfo or ChainInfoOwned

Buckram123: I think it's also possible to implement `TryFrom<ChainInfo> for Channel` using default `RUNTIME`

cw-orch-daemon/src/core.rs
```diff
@@ -66,34 +70,120 @@ pub const INSTANTIATE_2_TYPE_URL: &str = "/cosmwasm.wasm.v1.MsgInstantiateContra
...
     Use a Mutex on top of this DaemonAsync to avoid such errors.
 */
-pub struct DaemonAsyncBase<Sender: SenderTrait = Wallet> {
+pub struct DaemonAsyncBase<Sender = Wallet> {
     /// Sender to send transactions to the chain
-    pub sender: Sender,
+    pub(crate) sender: Sender,
     /// State of the daemon
-    pub state: DaemonState,
+    pub(crate) state: DaemonState,
```
Buckram123, line 77: Is there a special reason for this change? Some actions not possible without `sender` or `state`

cw-orch-daemon/src/queriers/cosmwasm.rs
```diff
@@ -19,15 +19,15 @@ use tonic::transport::Channel;
...
     pub channel: Channel,
     pub rt_handle: Option<Handle>,
     _sender: PhantomData<Sender>,
 }
 
-pub type CosmWasm = CosmWasmBase<Wallet>;
+pub type CosmWasm = CosmWasmBase<NoSender>;
 
-impl<Sender: SenderTrait> CosmWasmBase<Sender> {
+impl<Sender: QuerySender> CosmWasmBase<Sender> {
```
Buckram123, line 30: I love that NoSender is the default one here :rocket:

cw-orch-daemon/src/senders/no_sender.rs
```diff
@@ -1,63 +1,61 @@
...
-        })
+#[cfg(test)]
+mod tests {
+    use cw_orch_networks::networks::JUNO_1;
+
+    use crate::DaemonBuilder;
+
+    #[test]
+    fn build() {
+        let builder: QueryOnlyDaemon = DaemonBuilder::new(JUNO_1).build_sender(())?;
```
Buckram123, line 59: It's so clean :face_holding_back_tears:

Buckram123 (approved):
Great PR, left couple of nits!

cw-orch-daemon/src/senders/no_sender.rs
CyberHoward: It does always require specifying the return type of the function because there's no way to know which SenderOptions turn into which Senders...

Do you think it's worth adding/refactoring some code to make that possible?

Buckram123: > It does always require specifying the return type of the function because there's no way to know which SenderOptions turn into which Senders...
> 
> Do you think it's worth adding/refactoring some code to make that possible?

I think if we add PhantomData inside daemon builder to forward type to the constructed daemon and avoid specifying type, what do you think?

CyberHoward: Discussed removing `SenderOptions` type in favor of accepting `Sender` type itself in `fn sender_builder`.

CyberHoward: Actually this might not be that easy because the sender requires access to `ChainInfoOwned`, which is set in the daemon. 

I.e. if we remove senderoptions we could provide a sender that interacts with chain X while the daemon is configured for chain Y

cw-orch-daemon/src/channel.rs
CyberHoward: Edit: I mis-understood. That's a good idea!
I actually rarely use that trait. Is it good practice to use it?

Buckram123: > Edit: I mis-understood. That's a good idea! I actually rarely use that trait. Is it good practice to use it?

Yes from trait is usually good practice if you only need one type to get the other type. But is it bad pattern to use static runtime so mindlessly?

cw-orch-daemon/src/senders/cosmos.rs
```diff
@@ -29,297 +30,121 @@ use cosmrs::{
...
-            )
-        } else {
-            Sender::new_with_options(chain_info, grpc_channel, sender_options)
-        }
-    }
+    pub chain_info: Arc<ChainInfoOwned>,
+    pub(crate) options: CosmosOptions,
+    pub secp: Secp256k1<C>,
+    // Private field to ensure the struct is not constructible outside of this module
+    _private: (),
```
Buckram123, line 67:
```suggestion
    // Private field to ensure the struct is not constructible outside of this module
    _private: (),
```
options already private to this module

cw-orch-daemon/src/senders/tx.rs
```diff
@@ -0,0 +1,40 @@
...
+use cosmrs::{bank::MsgSend, tx::Msg, AccountId, Any};
+use cosmwasm_std::Addr;
+
+use crate::{parse_cw_coins, CosmTxResponse, DaemonError};
+
+use super::query::QuerySender;
+
+pub trait TxSender: QuerySender {
+    /// Returns the `AccountId` of the sender that commits the transaction.
+    fn account_id(&self) -> AccountId;
```
Buckram123, line 12: Does AccountId work with ethereum like addresses(`0x..`)? Just wonder, I'm not sure how good cw-orch with non-bech32 addresses work

Buckram123 (approved):
LGTM! There is many cool breaking changes :rocket:

cw-orch-daemon/src/senders/tx.rs
CyberHoward: i don't think it does. I'm not even sure if it should be in here tbh, don't know why nicolas added it...

cw-orch-daemon/src/channel.rs
CyberHoward: I don't think we should do that implicity. It's not really straight forward and i doubt that anyone will actually try to use this type manually.

cw-orch-daemon/src/senders/no_sender.rs
CyberHoward: Changed it so `SenderBuilder` is now a trait that's implemented on `SenderOptions`, so providing options allows for determining Sender type

CyberHoward:
Still need to update docs

PR #440, opened 2024-07-09
Bump cw-orch to 0.24.1

Buckram123:
### Checklist

- [x] Changelog updated.
- [ ] Docs updated.


Cargo.toml
```diff
@@ -35,30 +35,29 @@ anyhow = "1.0"
 serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 tokio = { version = "1.4", features = ["full"] }
 
-cw-orch = { path = "./cw-orch", version = "0.23.0" }
-cw-orch-daemon = { path = "./cw-orch-daemon", version = "0.23.2" }
-cw-orch-core = { path = "packages/cw-orch-core", version = "1.1.1" }
```
CyberHoward, line 40: I think because we added a "deprecated" flag, we should probably do a minor bump vs. a patch dump on core.

Buckram123: Already published with patch as we discussed yesterday. I think it's fine, as if doesn't break any existent code compilation, only few warnings

CyberHoward (approved):

PR #445, opened 2024-07-15
Add `is_test` for daemons to use temporary files 

Buckram123:
Randomly generated name for temporary file inside tempdir of OS will be used when `is_test` provided for daemons, this file will be removed by OS when no longer needed

### Checklist

- [x] Changelog updated.
- [ ] Docs updated.

### DIFF
--- a/cw-orch-daemon/src/state.rs
+++ b/cw-orch-daemon/src/state.rs
@@ -8,7 +8,6 @@ use cw_orch_core::{environment::StateInterface, log::local_target, CwEnvError};
 use once_cell::sync::Lazy;
 use serde::Serialize;
 use serde_json::{json, Value};
-use std::ffi::OsString;
 use std::sync::Arc;
 use std::{
     collections::{HashMap, HashSet},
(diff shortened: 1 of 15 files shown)


cw-orch-daemon/src/state.rs
```diff
@@ -331,6 +332,19 @@ impl StateInterface for DaemonState {
     }
 }
 
+// copied from `tempfile::util::tmpname` implementation
+pub(crate) fn gen_temp_file_path() -> std::path::PathBuf {
```
CyberHoward, line 336: We use uid and tempfile crates in node tests

https://docs.rs/uid/latest/uid/
https://docs.rs/tempfile/latest/tempfile/

Do you prefer this setup?

cw-orch-daemon/tests/daemon_state.rs
```diff
@@ -216,22 +216,22 @@ fn reuse_same_state_multichain() {
...
-            std::thread::sleep(std::time::Duration::from_millis(50));
-            let _state = JsonLockedState::new(TEST_STATE_FILE);
-        }
-        Err(_) => (),
-    }
-}
+// #[test]
+// #[serial_test::serial]
+// #[should_panic]
+// #[ignore = "Serial don't track forks for some reason, run it manually"]
```
CyberHoward, line 222: Can we update this test or why is it commented?

cw-orch-daemon/src/state.rs
Buckram123: Yeah uid is be nice, thanks!
I'm not sure if we can use tempfile crate as it returns `File` object, which means we don't have control of creation for this object, but we need it for locking file, beyond that it's pretty hacky to get path from the `File` object

Buckram123: Done, also noticed that uid not used (anymore?) in node tests, so removed unused part from there as well!

CyberHoward (approved):

PR #451, opened 2024-07-23
Network configs

Buckram123:
New feature network configs. It allows to customize chains passed inside Daemon Builder using a json file or generate new network configs.
To use this feature you will need a file `networks.json` located at `~/.cw-orchestrator/networks.json`. Format follows same structure as `ChainInfoOwned`. You can take template from `networks.json.example`

Code examples:
- Existing ChainInfo config:
  ```rust
    let daemon = Daemon::builder(cw_orch_daemon::networks::JUNO_1).build();
  ```
  ```json
  {"juno-1": {"gas_price": 0}}
  ```
  This will use juno config and replace gas price to 0

- Custom config
  ```rust
    let daemon = Daemon::builder(ChainInfo::config("abstr-1")).build();
  ```
  ```json
  {"abstr-1":{"kind":"mainnet","chain_id":"juno-1","gas_denom":"ujuno","gas_price":0.075,"grpc_urls":["http://juno-grpc.polkachu.com:12690"],"network_info":{"chain_name":"juno","pub_address_prefix":"juno","coin_type":118}}}
  ```
  This will use custom config for "abstr-1"

### Checklist

- [x] Changelog updated.
- [ ] Docs updated.


.github/workflows/test.yml
```diff
@@ -28,6 +28,11 @@ jobs:
       # https://twitter.com/jonhoo/status/1571290371124260865
       - name: cargo test --locked
         run: cargo test --locked --all-features --all-targets
+      - name: cargo test network config
```
CyberHoward, line 31: I don't think we need this test here if we have a unit-test that parses `networks.json.example`

cw-orch-daemon/tests/chain_info.rs
```diff
@@ -0,0 +1,114 @@
...
+    "gas_price": 0.075,
+    "grpc_urls": ["http://juno-grpc.polkachu.com:12690"],
+    "network_info": {
+      "chain_name": "juno",
+      "pub_address_prefix": "juno",
+      "coin_type": 118
+    }
+  }
+}
+***/
```
Buckram123, line 114: @CyberHoward Don't think we should put such mock in `networks.json.example`

packages/cw-orch-core/src/environment/chain_info.rs
```diff
@@ -32,8 +32,9 @@ pub struct ChainInfoBase<StringType: Into<String>, StringArrayType: AsRef<[Strin
 }
 
 /// Information about the underlying network, used for key derivation
-#[derive(Clone, Debug, Serialize, Default)]
-pub struct NetworkInfoBase<StringType> {
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+#[serde(default)]
+pub struct NetworkInfoBase<StringType: Into<String>> {
```
CyberHoward, line 37: Adding trait bounds is a breaking change.

cw-orch-daemon/tests/chain_info.rs
CyberHoward: Then we can save it as a seperate file in the `tests` folder. Having the CI push the file is something we need in other repos but we don't need to do it to test if the logic works. 
Should keep tests as simple as possible

CyberHoward:
Does this change make any changes to how the builder is used? 

How do I make use of this feature? (explain in PR description)

cw-orch-daemon/src/lib.rs
```diff
@@ -20,12 +20,14 @@ mod channel;
...
 mod error;
 mod log;
+mod network_config;
 mod state;
 mod sync;
 mod tx_resp;
 
 pub use self::{builder::*, channel::*, core::*, error::*, state::*, sync::*, tx_resp::*};
 pub use cw_orch_networks::networks;
+pub use network_config::NetworkConfig;
```
CyberHoward, line 30: would prefer to keep this type private

cw-orch-daemon/src/network_config.rs
```diff
@@ -0,0 +1,139 @@
...
+        };
+
+        match serde_json::from_slice(&data) {
+            Ok(network_config) => Some(Self(network_config)),
+            // Only this is critical
+            Err(e) => panic!("Invalid network config: {e}"),
+        }
+    }
+
+    pub fn update_chain_info(&self, mut chain_info: ChainInfoOwned) -> ChainInfoOwned {
```
CyberHoward, line 43: I'm a bit confused by this function. 

When would the chain_info object be empty?

A big part of this essentially merges two chain_info objects, should we not move that implementation to the ChainInfoOwned type and rename it 

`fn overwrite_with(self, chain_info) -> Self` or similar.

Buckram123: > I'm a bit confused by this function.
> 
> When would the chain_info object be empty?

`ChainInfo::config(chain_id)` creates empty chain info, except chain_id

> A big part of this essentially merges two chain_info objects, should we not move that implementation to the ChainInfoOwned type and rename it
> 
> `fn overwrite_with(self, chain_info) -> Self` or similar.

Ok!

cw-orch-daemon/src/network_config.rs
```diff
@@ -0,0 +1,96 @@
+use std::collections::HashMap;
+
+use cw_orch_core::environment::ChainInfoOwned;
+
+use crate::env::default_state_folder;
+
+pub(crate) fn load(chain_id: &str) -> Option<ChainInfoOwned> {
```
CyberHoward, line 7: Do you prefer having this be a seperate function over having it be a fn in the `ChainInfoOwned` implementation?

packages/cw-orch-core/Cargo.toml
```diff
@@ -18,9 +18,6 @@ all-features = true
 
 [features]
 default = []
-# enable node-backed tests (ensure Docker is running)
-# run with `cargo test --jobs 1 --features node-tests`
-node-tests = []
```
CyberHoward, line 23: can we comment them out instead of removing?

cw-orch-daemon/src/network_config.rs
Buckram123: > Do you prefer having this be a seperate function over having it be a fn in the `ChainInfoOwned` implementation?

I would like to keep it inside cw-orch-daemon **only** what do you think?

packages/cw-orch-core/Cargo.toml
Buckram123: This feature does nothing on cw-orch-core, any reason we would want to keep it?

cw-orch-daemon/src/network_config.rs
CyberHoward: Yeah that's good.

packages/cw-orch-core/Cargo.toml
CyberHoward: ahh oeps didn't see that. thanks!

CyberHoward (approved):
LGTM :)

PR #459, opened 2024-08-06
Starship improvements

Buckram123:
### Checklist

- [x] Changelog updated.
- [ ] Docs updated.


cw-orch-daemon/src/sync/builder.rs
```diff
@@ -128,6 +130,13 @@ impl DaemonBuilder {
         self
     }
 
+    /// Load network from `$HOME/.cw-orchestrator/networks.json`
```
CyberHoward, line 133: Do we want to change this to toml before merging?

CyberHoward (approved):
LGTM! Just want to see if we don't want to use toml instead of json for the network overwrite?

cw-orch-daemon/src/sync/builder.rs
Buckram123: Changed!

cw-orch-daemon/tests/mock_networks.toml
Buckram123, line 1: Thiis is 514 characters which is still lower than minimized json 531, so only benefits here

PR #390, opened 2024-05-13
Cosmwasm std 2 update

Kayanski:
This PR aims at updating cosmwas-std to 2.0 version.

I am blocked because Osmosis-std and test-tube need cosmwasm-std 1. 

- Use cw-multi-test branch for dev, don't rush it : `update/add-ibc-support-cosmwasm2`
- Use clone-cw-multi-test branch for dev, don't rush it : `update/local-support-cosmwasm2`

Gotta wait until osmosis test tube is separated OR osmosis-std updates

Temporary Ignored : 
- cw-orch-compatibility test
- Clone Testing (because clone-cw-multi-test needs cw-orch to be updated first)
- Osmosis Test tube (because osmosis-std hasn't updated)

PR #467, opened 2024-08-15
Cosmwasm2 accept addr

Buckram123:
Merge into #390
MockApi now accepts only valid Bech32 addrs, so it doesn't make sense to accept anything except Addr anymore

### Checklist

- [ ] Changelog updated.
- [ ] Docs updated.

### DIFF
--- a/contracts/mock_contract/src/lib.rs
+++ b/contracts/mock_contract/src/lib.rs
@@ -253,7 +253,7 @@ mod test {
 
         let cw2_info: cw2::ContractVersion = from_json(
             mock.wasm_querier()
-                .raw_query(contract.address()?, b"contract_info".to_vec())?,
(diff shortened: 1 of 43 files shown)


Buckram123:
Robin did mention that it's generally better to accept `Into<String>`, so another breaking thing to consider that we could do in this PR is replacing `impl ToString` to `impl Into<String>` (not for Addrs, but for path, client_id, etc..)

CyberHoward (approved):

CyberHoward:
> Robin did mention that it's generally better to accept `Into<String>` instead of `impl ToString`, so another breaking thing to consider that we could do in this PR is replacing to `impl Into<String>` (not for Addrs, but for path, client_id, etc..)

Can we do that in a different PR?

Issue #386, opened 2024-05-11
Bump CosmWasm to 2.0


CyberHoward:
We should bump our tooling to 2.0 as tooling should be one of the first packages that roll out with major version bumps in an ecosystem. 

See https://github.com/CosmWasm/sylvia/issues/374#issuecomment-2102845372.

Kayanski:
#390 solves this

Buckram123:
Bumped

PR #468, opened 2024-08-15
Separate contracts

Buckram123:
See ORC-152


packages/clone-testing/src/core.rs
```diff
@@ -519,7 +519,7 @@ mod test {
     impl Uploadable for MockCw20 {
         fn wasm(_chain: &ChainInfoOwned) -> WasmPath {
             let path = PathBuf::from_str(env!("CARGO_MANIFEST_DIR")).unwrap();
-            let path = path.join("../../artifacts/cw20_base.wasm");
+            let path = path.join("../../../artifacts/cw20_base.wasm");
```
CyberHoward, line 522: Are you sure about this change?[#386](https://github.com/AbstractSDK/cw-orchestrator/issues/386)
Is this even ever used?

wasm_contracts.sh
CyberHoward, line 1: let's move this to the contracts-ws dir

CyberHoward (approved):
Minor comments, LGTM!

packages/clone-testing/src/core.rs
Buckram123: Nice catch, it's not being used, we don't even have this file. I'll put `unimplemented!()` inside

wasm_contracts.sh
Buckram123: `artifacts/` still in the root of repo. Please check https://github.com/CosmWasm/optimizer?tab=readme-ov-file#contracts-excluded-from-workspace

CyberHoward:
Could you fix the CI before we merge?

CyberHoward:
@Buckram123 the links in the docs weren't update! 
image

PR #473, opened 2024-08-20
Flatten daemon state

Buckram123:
This PR flattens `.chain_name.chain_id` path on state files into `.chain_id`

### Checklist

- [x] Changelog updated.
- [x] Docs updated.


cw-orch/src/prelude.rs
```diff
@@ -60,6 +60,7 @@ pub use crate::daemon::{
     DaemonAsyncBuilder,
     // expose the sync variants
     DaemonBuilder,
+    DeployedChains,
```
CyberHoward, line 63: not sure this needs to be in prelude. Is it available through regular mod?

Buckram123: Yep, thanks!

CyberHoward (approved):

PR #480, opened 2024-09-03
Update to Cosmwasm-std 2.1 for Clone Testing

Kayanski:
This PR aims at upgrading cosmwasm-std to 2.1 inside clone testing.

This depends on https://github.com/AbstractSDK/cw-multi-test-fork/pull/7

**Dependencies**
- Clone-Testing now depends on the published version of cw-orch for compatibility reasons, because clone-testing depends on cw-multi-test-clone-testing which in turns depends on cw-orch. So we use the published version inside the dependencies and use a patch inside the workspace to make sure everything is compatible.

### Checklist

- [ ] Changelog updated.
- [ ] Docs updated.


packages/clone-testing/Cargo.toml
```diff
@@ -27,10 +32,12 @@ tonic = { workspace = true }
 [dev-dependencies]
 cosmwasm-schema = "2.0.0"
 speculoos = { workspace = true }
-cw20 = { workspace = true }
-cw20-base = { workspace = true }
+
+cw20 = { package = "cw20", version = "2.0.0" }
+cw20-base = { package = "cw20-base", version = "2.0.0" }
```
Buckram123, line 37:
```suggestion
cw20 = { version = "2.0.0" }
cw20-base = { version = "2.0.0" }
```

Buckram123 (approved):
small nit, LGTM

packages/clone-testing/Cargo.toml
```diff
@@ -11,10 +11,15 @@ description = "Adapter for interacting with cw-multi-test via the cw-orchestrato
 
 [dependencies]
 cosmwasm-std = { workspace = true }
-cw-orch-daemon = { workspace = true }
-cw-orch-core = { workspace = true }
-cw-orch-mock = { workspace = true }
-clone-cw-multi-test = { version = "0.5", features = ["cosmwasm_1_2"] }
+
+# This package depends on published cw-orch version due to the clone-testing-cw-multi-test dependency
+cw-orch-core = { version = "2.0.0" }
```
CyberHoward, line 16: Can we keep these imports as workspace/path imports?

Cargo.toml
```diff
@@ -89,3 +88,6 @@ log = "0.4.22"
 # Interchain
 ibc-relayer-types = { version = "0.29.2" }
 ibc-chain-registry = { version = "0.29.2" }
+
+[patch.crates-io]
+cw-orch-core = { path = "packages/cw-orch-core" }
```
CyberHoward, line 93: And remove this patch?

packages/clone-testing/Cargo.toml
Kayanski: No we can't unfortunately, because of the comment

Kayanski: The exact code where this fails is here : https://github.com/AbstractSDK/cw-orchestrator/blob/221391fb8b8460cdae6aa8333f99dc2f18c95ad8/packages/clone-testing/src/core.rs#L227
The `RemoteChannel` object takes in a ChainData type

CyberHoward: The comment says why the dependency is there, not why it's a version vs a path dependency.

CyberHoward: Having both the path **and** the version allows us to: 

- Publish (because then the version is used)
- Keep it up-to-date as dev-builds use the local version

So my question is why not use the path here, in addition to the version?

CyberHoward: And remove the patch in the process.

Kayanski: I re-iterate my point above. It's not a question of publishing, it's a question of dependencies. Without even tryong to publish, cargo check doesn't pass if the cw-orch-core version used here doesn't correspond to the version used in `clone-cw-multi-test`

packages/clone-testing/Cargo.toml
```diff
@@ -11,10 +11,16 @@ description = "Adapter for interacting with cw-multi-test via the cw-orchestrato
 
 [dependencies]
 cosmwasm-std = { workspace = true }
-cw-orch-daemon = { workspace = true }
-cw-orch-core = { workspace = true }
-cw-orch-mock = { workspace = true }
-clone-cw-multi-test = { version = "0.5", features = ["cosmwasm_1_2"] }
+
+# This package depends on published cw-orch version due to the clone-testing-cw-multi-test dependency
```
CyberHoward, line 15:
```suggestion
# This package depends on published cw-orch version due to the clone-testing-cw-multi-test dependency depending on the published version.
```

packages/clone-testing/Cargo.toml
```diff
@@ -27,10 +33,12 @@ tonic = { workspace = true }
...
-cw20 = { workspace = true }
-cw20-base = { workspace = true }
+
+cw20 = { version = "2.0.0" }
+cw20-base = { version = "2.0.0" }
 
 env_logger = "0.10.0"
 counter-contract = { path = "../../contracts-ws/contracts/counter" }
-cw-orch = { path = "../../cw-orch", features = ["daemon"] }
+# This package depends on published cw-orch version due to the clone-testing-cw-multi-test dependency
```
CyberHoward, line 42:
```suggestion
# This package depends on published cw-orch version due to the clone-testing-cw-multi-test dependency depending on the published version.
```

packages/clone-testing/Cargo.toml
```diff
@@ -13,9 +13,9 @@ description = "Adapter for interacting with cw-multi-test via the cw-orchestrato
 cosmwasm-std = { workspace = true }
 
 # This package depends on published cw-orch version due to the clone-testing-cw-multi-test dependency
-cw-orch-core = { version = "2.0.0" }
-cw-orch-mock = { version = "0.24.0" }
-cw-orch-daemon = { version = "0.25.0" }
+cw-orch-core = { path = "../cw-orch-core" }
```
CyberHoward, line 15: should be able to use `workspace = true` no? 

Otherwise you'll have trouble publishing this package.

packages/clone-testing/Cargo.toml
```diff
@@ -27,8 +32,9 @@ tonic = { workspace = true }
 [dev-dependencies]
 cosmwasm-schema = "2.0.0"
 speculoos = { workspace = true }
-cw20 = { workspace = true }
-cw20-base = { workspace = true }
+
+cw20 = { version = "2.0.0" }
```
CyberHoward, line 36: return this to `workspace = true`?

CyberHoward (approved):

packages/clone-testing/Cargo.toml
Kayanski: No, there is a version mismatch with cosmwasm inside abstract-cw20 repo. We'll need to change that as soon as the new interfaces are in (and potentially inside this repo)

packages/clone-testing/Cargo.toml
Kayanski: Yep

PR #481, opened 2024-09-04
Re-use contract's WASM compilation

Kayanski:
This PR introduces caching of wasm compiled modules to speed up the execution.

When using a wasm file, we compile it and save it inside the app for later use. 

This is possible everywhere except during in-contract queries. When using a contract inside an in-contract query, it doesn't get saved into memory.

Depends on (where all the changes are located) : https://github.com/AbstractSDK/cw-multi-test-fork/pull/8

### Checklist

- [ ] Changelog updated.
- [ ] Docs updated.

### DIFF
--- a/packages/clone-testing/Cargo.toml
+++ b/packages/clone-testing/Cargo.toml
@@ -17,7 +17,8 @@ cw-orch-core = { version = "2.0.0" }
 cw-orch-mock = { version = "0.24.0" }
 cw-orch-daemon = { version = "0.25.0" }
 
-clone-cw-multi-test = { version = "0.6.0-alpha.1" }
+# clone-cw-multi-test = { version = "0.6.0-alpha.1" }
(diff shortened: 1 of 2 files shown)


packages/clone-testing/Cargo.toml
```diff
@@ -11,10 +11,16 @@ description = "Adapter for interacting with cw-multi-test via the cw-orchestrato
...
 [dependencies]
 cosmwasm-std = { workspace = true }
-cw-orch-daemon = { workspace = true }
-cw-orch-core = { workspace = true }
-cw-orch-mock = { workspace = true }
-clone-cw-multi-test = { version = "0.5", features = ["cosmwasm_1_2"] }
+
+# This package depends on published cw-orch version due to the clone-testing-cw-multi-test dependency
+cw-orch-core = { version = "2.0.0" }
+cw-orch-mock = { version = "0.24.0" }
```
CyberHoward, line 17: same question as #480

Kayanski: Resolved because this branch will be merged into #480

PR #482, opened 2024-09-04
Allow setting private key for sender

Kayanski:
This PR aims at allowing setting the private key or the mnemonic for the CosmosSender object

### Checklist

- [x] Changelog updated.
- [x] Docs updated.

### DIFF
--- a/cw-orch-daemon/src/senders/cosmos.rs
+++ b/cw-orch-daemon/src/senders/cosmos.rs
@@ -149,7 +149,7 @@ impl Wallet {
     }
 
     /// Replaces the private key the sender is using
-    /// If you only need to change the mnemonic, use [Self::set_mnemonic]
+    /// You can use a mnemonic to overwrite the key using [Self::set_mnemonic]
     pub fn set_private_key(&mut self, private_key: PrivateKey) {
         self.private_key = private_key
     }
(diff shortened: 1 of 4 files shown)


cw-orch-daemon/src/senders/cosmos.rs
```diff
@@ -132,6 +132,28 @@ impl Wallet {
...
+            0,
+            self.options.hd_index.unwrap_or(0),
+            self.chain_info.network_info.coin_type,
+        )?;
+        self.set_private_key(pk);
+        Ok(())
+    }
+
+    /// Replaces the private key the sender is using
+    /// If you only need to change the mnemonic, use [Self::set_mnemonic]
```
CyberHoward, line 152: Changing the mnemonic = changing the private key. 
```suggestion
    /// You can use a mnemonic to overwrite the key using [Self::set_mnemonic]
```

cw-orch-daemon/src/senders/cosmos.rs
```diff
@@ -132,6 +132,28 @@ impl Wallet {
         self.options.clone()
     }
 
+    /// Replaces the private key that the [CosmosSender] is using with the provided mnemonic string.
```
CyberHoward, line 135:
```suggestion
    /// Replaces the private key that the [CosmosSender] is using with key derived from the provided 24-word mnemonic.
```

CyberHoward: And as a side-note: have we tried figuring out how to support 12-word mnemonics?

CyberHoward (approved):

PR #483, opened 2024-09-04
Update Docs for Daemon Builder and Sender

Kayanski:
This Pr aims at providing better and up-to-date documentation for Daemon and Builder Sender

### Checklist

- [x] Changelog updated.
- [x] Docs updated.

### DIFF
--- a/cw-orch-daemon/examples/hd_index.rs
+++ b/cw-orch-daemon/examples/hd_index.rs
@@ -1,0 +1,26 @@
+use cosmwasm_std::coins;
+use cw_orch::{anyhow, prelude::*};
+use cw_orch_daemon::CosmosOptions;
+
+// From https://github.com/CosmosContracts/juno/blob/32568dba828ff7783aea8cb5bb4b8b5832888255/docker/test-user.env#L1
+pub const LOCAL_JUNO_SENDER: &str = "juno16g2rahf5846rxzp3fwlswy08fz8ccuwk03k57y";
+pub const LOCAL_JUNO_GRANTER: &str = "juno1afhtjur8js4589xymu346ca7a5y5293xpuv6ry";
+
+pub fn main() -> anyhow::Result<()> {
+    pretty_env_logger::init(); // Used to log contract and chain interactions
+
+    let network = networks::LOCAL_JUNO;
+    // There is no need to register a mnemonic to use this daemon querier
+    let sender = CosmosOptions::default()
+        .hd_index(5)
+        .authz_granter(&Addr::unchecked(LOCAL_JUNO_GRANTER));
+    let chain = Daemon::builder(network).build_sender(sender)?;
+
+    chain.rt_handle.block_on(
+        chain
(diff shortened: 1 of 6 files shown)


cw-orch-daemon/examples/hd_index.rs
```diff
@@ -0,0 +1,26 @@
...
+
+// From https://github.com/CosmosContracts/juno/blob/32568dba828ff7783aea8cb5bb4b8b5832888255/docker/test-user.env#L1
+pub const LOCAL_JUNO_SENDER: &str = "juno16g2rahf5846rxzp3fwlswy08fz8ccuwk03k57y";
+pub const LOCAL_JUNO_GRANTER: &str = "juno1afhtjur8js4589xymu346ca7a5y5293xpuv6ry";
+
+pub fn main() -> anyhow::Result<()> {
+    pretty_env_logger::init(); // Used to log contract and chain interactions
+
+    let network = networks::LOCAL_JUNO;
+    // There is no need to register a mnemonic to use this daemon querier
```
CyberHoward, line 13: Was is meant here with "no need to register a mnemonic"? 

Do you mean that the mnemonic is read from the env variables?

docs/src/integrations/daemon.md
```diff
@@ -103,14 +103,28 @@ Here are the available options and fields you can use in the builder object:
...
+
+If you wish to use the default `CosmosSender` provided by default, you can use this simple pattern:
+
+```rust,ignore
+  let daemon = Daemon::builder(JUNO_1).build()?;
+```
+
+You can use the following functions on the `Daemon` object to customize some of its properties : 
+
+- `Daemon::authz_granter` allows you to use the authz module. If this method is used, the sender will send transactions wrapped inside an authz message sent by the specified `granter`. <a href="https://docs.cosmos.network/v0.46/modules/authz/" target="_blank">More info on the authz module</a>. <a href="https://docs.rs/cw-orch-daemon/latest/cw_orch_daemon/type.Daemon.html#method.authz_granter" target="_blank">Documentation Link</a>
```
CyberHoward, line 122: Honestly I'm not sure if exposing these functions is that helpful VS confusing for users. If it's already exposed on the sender, should we expose it on the daemon as well?

@Kayanski @Buckram123

CyberHoward (approved):

docs/src/integrations/daemon.md
Kayanski: It would look like this is the users code, that's bad for starting users but then again, this functionality is not necessarily for starters: 
```rust
daemon.sender_mut().set_authz_granter(granter);
```

CyberHoward: Yeah I do prefer that. 

Or they construct `CosmosSender` and configure it with authz before building with `build_sender(cosmos_sender)`

CyberHoward: Those two ways is already enough imo.

PR #484, opened 2024-09-05
Fix TLS issues

CyberHoward:
Fixes the TLS related issues that we've ran into for https URLs

### DIFF
--- a/cw-orch-daemon/Cargo.toml
+++ b/cw-orch-daemon/Cargo.toml
@@ -45,7 +45,7 @@ bech32 = { version = "0.11.0", default-features = false, features = ["alloc"] }
 hex = { version = "0.4.3" }
 ripemd = { version = "0.1.3" }
 tokio = { workspace = true, features = ["full"] }
-tonic = { workspace = true, features = ["tls", "tls-roots", "tls-native-roots"] }
+tonic = { workspace = true, features = ["tls"] }
 reqwest = { version = "0.12.5" }
 base64 = { version = "0.22.1" }
 hkd32 = { version = "0.7.0", features = ["bip39", "mnemonic", "bech32"] }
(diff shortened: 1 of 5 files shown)


cw-orch-daemon/Cargo.toml
```diff
@@ -45,7 +45,7 @@ bech32 = { version = "0.11.0", default-features = false, features = ["alloc"] }
 hex = { version = "0.4.3" }
 ripemd = { version = "0.1.3" }
 tokio = { workspace = true, features = ["full"] }
-tonic = { workspace = true, features = ["tls", "tls-roots"] }
+tonic = { workspace = true, features = ["tls"] }
```
Kayanski, line 48: Reallt removing the tls roots here ?

cw-orch-daemon/src/channel.rs
```diff
@@ -2,7 +2,8 @@ use cosmrs::proto::cosmos::base::tendermint::v1beta1::{
     service_client::ServiceClient, GetNodeInfoRequest,
 };
 use cw_orch_core::{environment::ChainInfoOwned, log::connectivity_target};
-use tonic::transport::{Channel, ClientTlsConfig};
+use http::Uri;
```
Kayanski, line 5: Have you tried this new code with `http://` prefixed URIs ?

Kayanski (approved):

cw-orch-daemon/Cargo.toml
CyberHoward: It works without so not sure why we should enable it.

cw-orch-daemon/src/channel.rs
CyberHoward: Yes, see the tests in `query_only.rs`. There's a http and https test there.

PR #485, opened 2024-09-09
Added upload_with_access

Kayanski:
This PR aims at allowing users to upload cosmwasm contracts with instantiate permission settings.

### Checklist

- [x] Changelog updated.
- [x] Docs updated.

### DIFF
--- a/cw-orch-daemon/src/core.rs
+++ b/cw-orch-daemon/src/core.rs
@@ -367,29 +367,7 @@ impl<Sender: TxSender> DaemonAsyncBase<Sender> {
         let store_msg = cosmrs::cosmwasm::MsgStoreCode {
             sender: self.sender().account_id(),
             wasm_byte_code,
-            instantiate_permission: access
-                .map(|a| {
-                    let response = match a {
-                        AccessConfig::Nobody => cosmrs::cosmwasm::AccessConfig {
-                            permission: cosmrs::cosmwasm::AccessType::Nobody,
-                            addresses: vec![],
-                        },
-                        AccessConfig::Everybody => cosmrs::cosmwasm::AccessConfig {
-                            permission: cosmrs::cosmwasm::AccessType::Everybody,
-                            addresses: vec![],
-                        },
-                        AccessConfig::AnyOfAddresses(addresses) => cosmrs::cosmwasm::AccessConfig {
-                            permission: cosmrs::cosmwasm::AccessType::AnyOfAddresses,
-                            addresses: addresses
-                                .into_iter()
-                                .map(|a| a.parse())
-                                .collect::<Result<_, _>>()?,
-                        },
-                        _ => unimplemented!(),
-                    };
-                    Ok::<_, DaemonError>(response)
-                })
-                .transpose()?,
+            instantiate_permission: access.map(access_config_to_cosmrs).transpose()?,
         };
 
         let result = self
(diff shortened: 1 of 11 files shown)


packages/cw-orch-core/src/environment/mod.rs
```diff
@@ -16,4 +16,4 @@ pub use queriers::{
     DefaultQueriers, Querier, QuerierGetter, QueryHandler,
 };
 pub use state::{ChainState, StateInterface};
-pub use tx_handler::{TxHandler, TxResponse};
+pub use tx_handler::{AccessConfig, TxHandler, TxResponse};
```
Buckram123, line 19: Are we sure about exposing this type? Don't think proto messages have good Rust API

Buckram123 (approved):
I guess it would be rarely used, so LGTM. But if we want it to have nice API, don't think it makes sense for `AccessType::Unspecified` `AcessType::Nobody` and `AcessType::Everybody` to accept `addresses: Vec<String>`

packages/cw-orch-core/src/environment/mod.rs
Kayanski: I wanted to expose it for users to be able to have it without having to import `cosmrs` themselves.
But having a better API with addresses as String instead of complicated types is also a good idea

CyberHoward: Does this mean that cosmrs is inside of cw-orch-core by default?

CyberHoward: nevermind, looks like cosmos-sdk-proto is already included there by default.

Kayanski: New version is out with a slightly better API, what do you think ?

packages/cw-orch-core/src/environment/tx_handler.rs
```diff
@@ -81,6 +88,35 @@ pub trait TxHandler: ChainState + Clone {
     }
 }
 
+#[non_exhaustive]
+pub enum AccessConfig {
+    Nobody,
+    Everybody,
+    AnyOfAddresses(Vec<String>),
+}
```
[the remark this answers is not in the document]
Kayanski: yes that's what I thought, and non-exhaustive if we add it later

cw-orch/src/prelude.rs
```diff
@@ -27,8 +27,9 @@ pub use crate::environment::IndexResponse;
 
 // Environment
 pub use crate::environment::{
-    BankQuerier, BankSetter, CwEnv, DefaultQueriers, EnvironmentInfo, EnvironmentQuerier,
-    NodeQuerier, QuerierGetter, QueryHandler, TxHandler, TxResponse, WasmQuerier,
+    AccessConfig, BankQuerier, BankSetter, CwEnv, DefaultQueriers, EnvironmentInfo,
```
CyberHoward, line 30: I don't think AccessConfig should be in predule.
```suggestion
    BankQuerier, BankSetter, CwEnv, DefaultQueriers, EnvironmentInfo,
```

packages/cw-orch-core/src/environment/tx_handler.rs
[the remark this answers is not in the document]
CyberHoward: Are we sure `Unspecified` is the default?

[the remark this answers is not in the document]
CyberHoward: or how it behaves vs the other ones. 

Leaving it out seems weird to me.

cw-orch-daemon/src/core.rs
```diff
@@ -356,7 +367,29 @@ impl<Sender: TxSender> DaemonAsyncBase<Sender> {
         let store_msg = cosmrs::cosmwasm::MsgStoreCode {
             sender: self.sender().account_id(),
             wasm_byte_code,
-            instantiate_permission: None,
+            instantiate_permission: access
+                .map(|a| {
+                    let response = match a {
```
CyberHoward, line 372: Move this into a `try_into` implementation?

PR #486, opened 2024-09-09
12 words mnemonic

Kayanski:
This PR adds support for 12 and 24 words mnemonics.

### Checklist

- [ ] Changelog updated.
- [ ] Docs updated.

### DIFF
--- a/cw-orch-daemon/src/keys/private.rs
+++ b/cw-orch-daemon/src/keys/private.rs
@@ -11,9 +11,9 @@ use bitcoin::{
 };
 use cosmrs::tx::SignerPublicKey;
 use cw_orch_core::log::local_target;
-use hkd32::mnemonic::{Phrase, Seed};
+use itertools::Itertools;
 use prost_types::Any;
-use rand_core::OsRng;
+use rand_core::{OsRng, RngCore};
 
 /// The Private key structure that is used to generate signatures and public keys
 /// WARNING: No Security Audit has been performed
@@ -25,8 +25,8 @@ pub struct PrivateKey {
     pub index: u32,
     #[allow(missing_docs)]
     pub coin_type: u32,
-    /// The 24 words used to generate this private key
-    mnemonic: Option<Phrase>,
+    /// The mnemonic (12, 15, 18, 21 or 24 words) used to generate this private key
+    mnemonic: Option<bip39::Mnemonic>,
     #[allow(dead_code)]
     /// This is used for testing
     root_private_key: Xpriv,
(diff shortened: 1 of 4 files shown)


cw-orch-daemon/src/keys/private.rs
```diff
@@ -39,19 +39,22 @@ impl PrivateKey {
...
-        let phrase = hkd32::mnemonic::Phrase::random(OsRng, hkd32::mnemonic::Language::English);
-
-        PrivateKey::gen_private_key_phrase(secp, phrase, 0, 0, coin_type, seed_phrase)
+        let mut key = [0u8; 16];
+        OsRng.fill_bytes(&mut key);
+        match bip39::Mnemonic::from_entropy(&key) {
+            Ok(mnemonic) => {
+                PrivateKey::gen_private_key_phrase(secp, mnemonic, 0, 0, coin_type, seed_phrase)
+            }
+            Err(_) => Err(DaemonError::Phrasing),
```
CyberHoward, line 56: How can this ever err if menemonic is always created from a random key?

cw-orch-daemon/src/keys/private.rs
```diff
@@ -168,15 +167,14 @@ impl PrivateKey {
     // Generate private key from Phrase
     fn gen_private_key_phrase<C: secp256k1::Signing + secp256k1::Context>(
         secp: &Secp256k1<C>,
-        phrase: Phrase,
+        phrase: bip39::Mnemonic,
         account: u32,
         index: u32,
         coin_type: u32,
         seed_phrase: &str,
```
CyberHoward, line 174: We're calling this the seed_phrase but it looks like it's the password that can be used to generate the private key, allongside the seed_phrase.

cw-orch-daemon/src/keys/private.rs
```diff
@@ -39,19 +39,22 @@ impl PrivateKey {
...
     pub fn new_seed<C: secp256k1::Signing + secp256k1::Context>(
         secp: &Secp256k1<C>,
         seed_phrase: &str,
         coin_type: u32,
     ) -> Result<PrivateKey, DaemonError> {
-        let phrase = hkd32::mnemonic::Phrase::random(OsRng, hkd32::mnemonic::Language::English);
-
-        PrivateKey::gen_private_key_phrase(secp, phrase, 0, 0, coin_type, seed_phrase)
+        let mut key = [0u8; 16];
+        OsRng.fill_bytes(&mut key);
```
Buckram123, line 51: Why not reuse built-in rng in bip39? https://docs.rs/bip39/2.0.0/src/bip39/lib.rs.html#290-293

Buckram123 (changes requested):
Let's use built-in rng from bip39

cw-orch-daemon/src/keys/private.rs
Kayanski: good catch, didn't find it

cw-orch-daemon/src/keys/private.rs
Kayanski: I don't know but it returns a Result, so I have to check that. 
If it ever errors, it will output `Error`  which is not very good information, so I prefer Daemon error

cw-orch-daemon/src/keys/private.rs
Kayanski: It's called a passphrase, so renaming the variable

PR #487, opened 2024-09-11
Added upload wasm function to sender

Kayanski:
This PR adds a `upload_wasm` function to `CosmosSender` for uploading wasm code associated to no Contract structure

### Checklist

- [x] Changelog updated.
- [ ] Docs updated.

### DIFF
--- a/cw-orch-daemon/src/core.rs
+++ b/cw-orch-daemon/src/core.rs
@@ -20,13 +20,11 @@ use cw_orch_core::{
     environment::{AsyncWasmQuerier, ChainInfoOwned, ChainState, IndexResponse, Querier},
     log::transaction_target,
 };
-use flate2::{write, Compression};
 use prost::Message;
 use serde::{de::DeserializeOwned, Serialize};
 use serde_json::from_str;
 use std::{
     fmt::Debug,
-    io::Write,
     ops::Deref,
     str::{from_utf8, FromStr},
     time::Duration,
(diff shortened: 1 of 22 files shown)


cw-orch-daemon/src/senders/cosmos.rs
```diff
@@ -415,6 +417,25 @@ impl Wallet {
...
+        let store_msg = cosmrs::cosmwasm::MsgStoreCode {
+            sender: self.account_id(),
+            wasm_byte_code,
+            instantiate_permission: None,
+        };
+
+        self.commit_tx(vec![store_msg], None).await
+    }
+}
+
```
Buckram123, line 438: Is it possible to move implementation from `DaemonAsyncBase::upload` here instead?

Buckram123: So there will be left something like:
```rust
let wasm_path = <T as Uploadable>::wasm(self.chain_info());
self.sender.upload_wasm(wasm_path)
```

Kayanski: Ok, added the `upload_wasm` function directly on the Sender trait, so it's not a CosmosSender specific thing anymore as well :)

cw-orch-daemon/src/senders/tx.rs
```diff
@@ -35,4 +39,27 @@ pub trait TxSender: QuerySender {
 
         self.commit_tx_any(msgs, memo)
     }
+
+    /// Uploads the `WasmPath` path specifier on chain.
+    /// The resulting code_id can be extracted from the Transaction result using [cw_orch_core::environment::IndexResponse::uploaded_code_id] and returns the resulting code_id
+    fn upload_wasm(
```
CyberHoward, line 45: I don't think this should be in this trait. Having it as a function on 
`impl CosmosSender` should be sufficient. 

An upload is just a specific case for a tx and will very unlikely be used with anything other than `CosmosSender`.

cw-orch-daemon/src/core.rs
```diff
@@ -349,32 +347,18 @@ impl<Sender: TxSender> DaemonAsyncBase<Sender> {
 
         log::debug!(target: &transaction_target(), "Uploading file at {:?}", wasm_path);
 
-        let file_contents = std::fs::read(wasm_path.path())?;
-        let mut e = write::GzEncoder::new(Vec::new(), Compression::default());
```
CyberHoward, line 353: So think this code should be left untouched as well.

CyberHoward (changes requested):
I'd like this function to just be added to the `CosmosSender` instead of being included in the core sender traits.

cw-orch-daemon/src/senders/tx.rs
Kayanski: Ok, reverting [6b0742c](https://github.com/AbstractSDK/cw-orchestrator/pull/487/commits/6b0742ce0aa5860dda9cb8a40a7e47f466cdceb6) and creating a function to avoid code duplication

cw-orch-daemon/src/core.rs
Kayanski: Ok, reverting [6b0742c](https://github.com/AbstractSDK/cw-orchestrator/pull/487/commits/6b0742ce0aa5860dda9cb8a40a7e47f466cdceb6) and creating a function to avoid code duplication

Buckram123 (approved):
LGTM

cw-orch-daemon/src/senders/cosmos.rs
```diff
@@ -415,6 +418,44 @@ impl Wallet {
...
+    pub async fn upload_with_access_config(
+        &self,
+        wasm_path: WasmPath,
+        access: Option<AccessConfig>,
+    ) -> Result<CosmTxResponse, DaemonError> {
+        upload_wasm(self, wasm_path, access).await
+    }
+}
+
+pub async fn upload_wasm<T: TxSender>(
```
CyberHoward, line 438: This should be in `core.rs` imo.

PR #489, opened 2024-09-17
Add distant wasm upload

Kayanski:
This PR aims at providing integrators with more methods to query the wasms to upload on chain.

### Checklist

- [x] Changelog updated.
- [ ] Docs updated.


packages/cw-orch-core/src/contract/paths/artifacts_dir.rs
```diff
@@ -0,0 +1,183 @@
+const ARM_POSTFIX: &str = "-aarch64";
```
Kayanski, line 1: This is just a copy of the old artifacts_dir module inside wasm_path.rs

CyberHoward (commented):
I'm surprised by how much code was changed for this feature, and it's breaking to cw-orch-core. 

is there no alternative way that we could structure this PR to make it non-breaking?

Kayanski:
> I'm surprised by how much code was changed for this feature, and it's breaking to cw-orch-core.
> 
> is there no alternative way that we could structure this PR to make it non-breaking?

Yes. this was done to propose a preliminary implementation. 
The issue here is that the current cw-orch-core version exposes WasmPath(PathBuf), which we need to modify to accomodate for this change. 
An additional option we could use is to have an additional method inside the `Uploadable` trait for fetching the contracts remotely. This is not ideal, but could be experimented. 
Removing the associated cw-plus changes to make things clearer

CyberHoward:
And what about separating the fetching from the rest of the logic? I.e. adding a fn or object that you can use to fetch any wasm blob and store it in the local / configured artifacts folder?

Kayanski:
This PR is breaking to cw-orch-core and won't be merged anytime soon. 
Keeping it as an Issue for tracking purposes

PR #491, opened 2024-09-18
Replace Cw-Multi-Test wiith official version

Kayanski:
This PR aims at removing the dependency of our fork of cw-multi-test and moving it to the official cw-multi-test.

This is still a draft because we are waiting for cw-multi-test to stabilize version 3.0.0

### Checklist

- [ ] Changelog updated.
- [ ] Docs updated.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,9 +33,6 @@ cosmwasm-std = { version = "2.1" }
 cw-storage-plus = { version = "2.0.0" }
 cosmos-sdk-proto = { version = "0.24.0", default-features = false }
 
-# cw-multi-test = { package = "abstract-cw-multi-test", version = "2.0.2", features = [
-#   "cosmwasm_1_2",
-# ] }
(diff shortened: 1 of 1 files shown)


Buckram123:
By the way, are we supposed duplicate cosmwasm set of features like cw-multi-test do? 
https://github.com/CosmWasm/cw-multi-test/blob/b846f156f72e98d2037710dab856544fbb3a49db/Cargo.toml#L17-L27
Or setting latest version features on cosmwasm packages would be sufficient?

Kayanski:
> By the way, are we supposed duplicate cosmwasm set of features like cw-multi-test do? https://github.com/CosmWasm/cw-multi-test/blob/b846f156f72e98d2037710dab856544fbb3a49db/Cargo.toml#L17-L27 Or setting latest version features on cosmwasm packages would be sufficient?

I'm not sure. setting the last one should be sufficient. 
Note here that when execution 2.0, all the features are available by default so we don't need to bother. If a user needs the specific features, we might think of that, but definitely not right now. I haven't heard someone complain about this

PR #492, opened 2024-09-18
Even better API for InterchainEnv

Kayanski:
This PR aims at providing a better API to InterchainEnv for all its purposes

 Checklist

- [ ] Changelog updated.
- [ ] Docs updated.

### DIFF
--- a/packages/interchain/interchain-core/src/env.rs
+++ b/packages/interchain/interchain-core/src/env.rs
@@ -109,7 +109,7 @@ pub type ChainId<'a> = &'a str;
 /// // This makes sure that the packets arrive successfully and present a success ack
 /// let result = interchain.await_and_check_packets("juno-1", tx_resp).unwrap();
 /// ```    
-pub trait InterchainEnv<Chain: IbcQueryHandler> {
+pub trait InterchainEnv<Chain: IbcQueryHandler>: Clone {
     /// Type returned by the internal channel creation function
     /// Examples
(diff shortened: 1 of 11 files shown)


CyberHoward (approved):

packages/interchain/interchain-core/src/env.rs
```diff
@@ -138,6 +138,19 @@ pub trait InterchainEnv<Chain: IbcQueryHandler> {
...
+    /// use cw_orch::prelude::*;
+    /// use cw_orch_interchain::prelude::*;
+    /// use counter_contract::CounterContract;
+    /// let interchain = MockBech32InterchainEnv::new(vec![("osmosis-1","osmo"),("archway-1","arch")]);
+    ///
+    /// let all_chains = interchain.chains().unwrap();
+    ///
+    /// ```
+    fn chains(&self) -> Result<Vec<Chain>, Self::Error>;
+
```
Buckram123, line 153: Can't we return iterator instead?

Buckram123: And why it returns Result, at what use-case it can error?

Buckram123 (changes requested):

packages/interchain/interchain-core/src/env.rs
Kayanski: Yes, no Error here, I don't see why it could err. And returning iterator to chain reference as well here

PR #494, opened 2024-09-26
Added CosmosSigner default implementation

Kayanski:
This PR aims at making the TxBroadcaster available to other types of Signers that want to implement their own signing logic, very close to the CosmosSender signing logic. 

This implementation is done to accomodate for Xion Senders that have a different signing mechanism

### Checklist

- [ ] Changelog updated.
- [ ] Docs updated.

### DIFF
--- a/cw-orch-daemon/src/senders/sign.rs
+++ b/cw-orch-daemon/src/senders/sign.rs
@@ -1,5 +1,4 @@
 use crate::{
-    cosmos_modules,
     queriers::Node,
     tx_broadcaster::{
         account_sequence_strategy, assert_broadcast_code_cosm_response, insufficient_fee_strategy,
@@ -20,7 +19,7 @@ pub struct SigningAccount {
     pub sequence: u64,
 }
 
-pub trait CosmosSigner: QuerySender<Error = DaemonError> + Sync {
+pub trait Signer: QuerySender<Error = DaemonError> + Sync {
     // --- General information about the signer --- //
     /// The chain id of the connected chain
     fn chain_id(&self) -> String;
@@ -75,7 +74,7 @@ pub trait CosmosSigner: QuerySender<Error = DaemonError> + Sync {
     }
 }
 
-impl<T: CosmosSigner + Sync> TxSender for T {
+impl<T: Signer + Sync> TxSender for T {
     fn account_id(&self) -> cosmrs::AccountId {
         self._account_id()
     }
--- a/cw-orch-daemon/src/senders/sign.rs
+++ b/cw-orch-daemon/src/senders/sign.rs
@@ -73,26 +73,6 @@ pub trait CosmosSigner: QuerySender<Error = DaemonError> + Sync {
                 .await
         }
     }
-    /// Transaction broadcasting for Tendermint Transactions
-    fn broadcast_tx(
-        &self,
-        tx: Raw,
-    ) -> impl std::future::Future<
-        Output = Result<cosmrs::proto::cosmos::base::abci::v1beta1::TxResponse, DaemonError>,
-    > + Send {
-        async move {
-            let mut client = cosmos_modules::tx::service_client::ServiceClient::new(self.channel());
-            let commit = client
-                .broadcast_tx(cosmos_modules::tx::BroadcastTxRequest {
-                    tx_bytes: tx.to_bytes()?,
-                    mode: cosmos_modules::tx::BroadcastMode::Sync.into(),
-                })
-                .await?;
-
-            let commit = commit.into_inner().tx_response.unwrap();
-            Ok(commit)
-        }
-    }
 }
 
 impl<T: CosmosSigner + Sync> TxSender for T {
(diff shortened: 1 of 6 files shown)


cw-orch-daemon/src/senders/sign.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+pub trait CosmosSigner: QuerySender<Error = DaemonError> + Sync {
+    // --- General information about the signer --- //
+    /// The chain id of the connected chain
+    fn chain_id(&self) -> String;
+
+    /// The account id of the signer.
+    /// This is `_` prefixed because this conflict wih the TxSender trat
+    fn _account_id(&self) -> AccountId;
```
CyberHoward, line 30: If it's a signer, it should also be a `TxSender` no? Can we add `TxSender` to the trait bound of this type to prevent this conflict?

CyberHoward: Or alternatively, remove `account_id` from `TxSender` and make `TxSender: CosmosSigner`

cw-orch-daemon/src/senders/sign.rs
```diff
@@ -0,0 +1,145 @@
...
+    AccountId, Any,
+};
+use prost::Message;
+
+pub struct SigningAccount {
+    pub account_number: u64,
+    pub sequence: u64,
+}
+
+pub trait CosmosSigner: QuerySender<Error = DaemonError> + Sync {
```
CyberHoward, line 23: Can we just call it `Signer`?

cw-orch-daemon/src/senders/sign.rs
```diff
@@ -0,0 +1,145 @@
...
+                })
+                .await?;
+
+            let commit = commit.into_inner().tx_response.unwrap();
+            Ok(commit)
+        }
+    }
+}
+
+impl<T: CosmosSigner + Sync> TxSender for T {
```
CyberHoward, line 98: It looks like having the trait bound `TxSender: Signer` makes sense!

Kayanski: A TxSender can be anything that is used to interact with a Daemon. It could be a MultisigSender (no signing capabilities), it could be a QuerySender.. So Being a Signer is not a prerequisite for being a sender. However being a Signer makes you a sender

cw-orch-daemon/src/tx_broadcaster.rs
```diff
@@ -67,7 +68,7 @@ impl TxBroadcaster {
     pub async fn broadcast(
         mut self,
         mut tx_builder: TxBuilder,
-        wallet: &Wallet,
+        wallet: &impl Signer,
```
CyberHoward, line 71:
```suggestion
        signer: &impl Signer,
```

cw-orch-daemon/src/tx_broadcaster.rs
```diff
@@ -122,7 +123,7 @@ fn strategy_condition_met(
 
 async fn broadcast_helper(
     tx_builder: &mut TxBuilder,
-    wallet: &Wallet,
+    wallet: &impl Signer,
```
CyberHoward, line 126:
```suggestion
    signer: &impl Signer,
```

cw-orch-daemon/src/tx_builder.rs
```diff
@@ -77,13 +80,12 @@ impl TxBuilder {
     }
 
     /// Simulates the transaction and returns the necessary gas fee returned by the simulation on a node
-    pub async fn simulate(&self, wallet: &Wallet) -> Result<u64, DaemonError> {
+    pub async fn simulate(&self, wallet: &impl Signer) -> Result<u64, DaemonError> {
```
CyberHoward, line 83:
```suggestion
    pub async fn simulate(&self, signer: &impl Signer) -> Result<u64, DaemonError> {
```

cw-orch-daemon/src/senders/sign.rs
CyberHoward: MultisigSender would still need a key to do signing.

CyberHoward: Otherwise it's impossible to vote or do proposals

cw-orch-daemon/src/senders/sign.rs
CyberHoward: If we don't rename it or move it then we can still call it `account_id`. 

If both traits are implemented the call just needs to specify the trait. 

`Signer::account_id`
`TxSender::account_id`

CyberHoward (approved):

Kayanski:
@CyberHoward had to modify the trait slightly to be able to merge with main and the corrected authz capabilities

cw-orch-daemon/src/senders/sign.rs
```diff
@@ -0,0 +1,136 @@
...
+    pub sequence: u64,
+}
+
+pub trait Signer: QuerySender<Error = DaemonError> + Sync {
+    // --- General information about the signer --- //
+    /// The chain id of the connected chain
+    fn chain_id(&self) -> String;
+
+    /// The account id of the signer.
+    /// This is `_` prefixed because this conflict wih the TxSender trait
```
CyberHoward, line 29:
```suggestion
```

PR #495, opened 2024-09-26
Correct authz capabilities

Kayanski:
Authz capabilities are wrong inside cw-orch-daemon when interacting with contracts. This PR aims to correct that

### Checklist

- [ ] Changelog updated.
- [ ] Docs updated.

### DIFF
--- a/cw-orch-daemon/src/core.rs
+++ b/cw-orch-daemon/src/core.rs
@@ -234,7 +234,7 @@ impl<Sender: TxSender> DaemonAsyncBase<Sender> {
         contract_address: &Addr,
     ) -> Result<CosmTxResponse, DaemonError> {
         let exec_msg: MsgExecuteContract = MsgExecuteContract {
-            sender: self.sender().account_id(),
+            sender: self.sender().msg_sender().map_err(Into::into)?,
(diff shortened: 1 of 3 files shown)


cw-orch-daemon/src/core.rs
```diff
@@ -380,7 +380,7 @@ pub async fn upload_wasm<T: TxSender>(
     e.write_all(&file_contents)?;
     let wasm_byte_code = e.finish()?;
     let store_msg = cosmrs::cosmwasm::MsgStoreCode {
-        sender: sender.account_id(),
+        sender: sender.msg_sender().map_err(Into::into)?,
```
CyberHoward, line 383: I don't think we need to authz for uploads

Kayanski: why not ?

CyberHoward: Hmm interesting

Kayanski: I am actually using it currently for Stargaze uploads :)

CyberHoward: cool! then LGMT :)

PR #501, opened 2024-09-30
Remove polytone dep

Buckram123:
We only use 1 type from polytone but it comes with all cosmwasm v1 packages, we should temporary replace it


Kayanski (approved):
Yes, very good idea ! If this removes cosmwasm-std 1 all-together, LGTM

Buckram123:
> Yes, very good idea ! If this removes cosmwasm-std 1 all-together, LGTM

Let's merge it with #500 and see

PR #503, opened 2024-09-30
Refactor/interchain core

Kayanski:
This PR aims at refactoring interchain-core by providing better types and abstractions, namely brings a separation between results (following the packet flow) and analysis (making sure there is no timeout and no ack errors).

TODO (maybe in another PR)
--> Add the possibility to error on error during acknowledgement execution

### Checklist

- [ ] Changelog updated.
- [ ] Docs updated.


packages/interchain/interchain-core/src/packet.rs
```diff
@@ -0,0 +1,335 @@
...
+        ack_tx: T,
+        /// The raw binary acknowledgement retrieved from `ack_tx`
+        ack: Binary,
+    },
+}
+
+/// The result of following a single packet across IBC
+/// This allows indentifying the different transactions involved as well as the result of the packet transmission
+#[derive(Clone)]
+#[must_use = "We recommend using `PacketAnalysis::into_result()` to assert ibc success"]
```
CyberHoward, line 48: Update comment

packages/interchain/interchain-core/src/packet.rs
```diff
@@ -0,0 +1,335 @@
...
+    pub send_tx: Option<TxId<Chain>>,
+    /// Outcome transactions of the packet (+ eventual acknowledgment)
+    pub outcome: IbcPacketOutcome<TxId<Chain>>,
+}
+
+/// The result of following all packets sent across IBC during a single transaction.
+///
+/// This structure is nested and will also await all packets emitted during the subsequent (receive/acknowledgement/timeout) transactions
+#[derive(Clone)]
+#[must_use = "We recommend using `PacketAnalysis::into_result()` to assert ibc success"]
```
CyberHoward, line 63: Update comment

packages/interchain/interchain-core/src/packet.rs
```diff
@@ -0,0 +1,335 @@
...
+                SuccessfullAck::Ics004(vec) => SuccessfullAck::Ics004(vec),
+                SuccessfullAck::Custom(_) => unreachable!(),
+            }
+        }
+    }
+
+    /// Success packet outcome. This is the result of a packet analysis.
+    /// The T generic is used to allow for raw transactions or analyzed transactions to be used
+    #[derive(Debug, PartialEq, Clone)]
+    pub struct SuccessIbcPacketOutcome<T, CustomOutcome = Empty> {
```
CyberHoward, line 251: I'm a bit confused on what is meant with "success". Does it mean that the ibc packet got acknowledged? On the topic of naming I think we should also use `Result` over `Outcome`, or maybe `Status` as it's only really a result after it's asserted.

What this datatype wants to reflect is a packet that has succesfully gone through its lifecycle. So mabye `AcknowledgedIbcPacket` would be even better?

packages/interchain/interchain-core/src/packet.rs
```diff
@@ -0,0 +1,335 @@
...
+    pub sequence: Sequence,
+    /// Chain identification to which the packet was sent
+    pub dst_chain_id: NetworkId,
+}
+
+/// Raw packet outcome
+/// The T generic is used to allow for raw transactions or analyzed transactions to be used
+#[derive(Debug, PartialEq, Clone)]
+#[must_use = "We recommend using `PacketAnalysis::into_result()` to assert ibc success"]
+pub enum IbcPacketOutcome<T> {
```
CyberHoward, line 28: What's the difference between this type and `SuccessIbcPacketOutcome`? 

It seems like the `Success` variants should be `SuccessIbcPacketOutcome`?

CyberHoward (changes requested):
Some comments on naming

packages/interchain/interchain-core/src/packet.rs
Kayanski: Success is not very well defined for now, but it should be very well defined on this crate. For now, usage has shown that success necessitates at least : 
- Packet has been acknowledged
- Acknowledgement has been successfully parsed and the parsed ack corresponds to a successful IBC operation

We could also add : 
- Acknowledgment execution has been successful (though it's way more difficult to know because ack transactions can't fail). 

So it's not a result (which in Rust semantics has 2 possible outcomes), it's an Result::Ok, if you will, so a Success here. 
- AcknowledgedIbcPacket is a little limited considering the above
- Status is also limited

packages/interchain/interchain-core/src/packet.rs
Kayanski: IbcPacketOutcome is the raw outcome from the packet cycle awaiting. It tells you what the packet lifecycle looks like for this packet. 
The Success variant corresponds to lifecycle with an ack. But not to a successful ack, because nothing has been parsed yet

packages/interchain/interchain-core/src/packet.rs
CyberHoward: > Acknowledgement has been successfully parsed and the parsed ack corresponds to a successful IBC operation

Does it make sense to have a separate step between `RawAck` and `AssertedAck`? 

I.e. why would we want a type that we know is a packet that went through its cycle and is of a specific datastructure but that we don't know the result of? 

Can we omit the type and simply talk about acknowledgements that were committed (no guarantees on the formatting or result of the packet) and then a checked type that does the whole parsing and checking into a result?

packages/interchain/interchain-core/src/packet.rs
CyberHoward: I would like us to differentiate between the low-level IBC lifecycle and whatever lifecycle the application expects. 

A packet either has a timeout or is acknowledged. Having an acknowledged packet doesn't give any guarantees about the data with that ack, and that is OK but we should be very precise with our names. 

I propose we use the IBC-native naming conventions to talk about the packet lifecycle (send, commit, ack, timeout) and we don't reuse those terms when talking about the other assertions that could be done on the data within the ack. 

We can then also get rid of terms like "success". If the packet is acknowledged then that's enough to proceed to the next step in the flow.

packages/interchain/interchain-core/src/packet.rs
```diff
@@ -0,0 +1,343 @@
...
+    /// Success packet outcome. This is the result of a packet analysis.
+    /// The T generic is used to allow for raw transactions or analyzed transactions to be used
+    #[derive(Debug, PartialEq, Clone)]
+    pub struct SuccessIbcPacketOutcome<T, CustomOutcome = Empty> {
+        /// The packets gets transmitted to the dst chain
+        pub receive_tx: T,
+        /// The ack is broadcasted back on the src chain
+        pub ack_tx: T,
+        /// The raw binary acknowledgement retrieved from `ack_tx`
+        pub ack: SuccessfullAck<CustomOutcome>,
```
CyberHoward, line 265:
```suggestion
        pub ibc_app_result: IbcAppResult<CustomResult>,
```

PR #508, opened 2024-10-14
Fixes: set_contracts_state

Buckram123:
Small fix to ensure state is patched(also refactored confusing nesting) and that we don't try to use chain_name in path

### DIFF
--- a/cw-orch-daemon/src/state.rs
+++ b/cw-orch-daemon/src/state.rs
@@ -357,20 +357,13 @@ pub trait DeployedChains<Chain: CwEnv>: cw_orch_core::contract::Deploy<Chain> {
     /// Set the default contract state for a contract, so that users can retrieve it in their application when importing the library
     /// If a state is provided, it is used for all contracts, otherwise, the state is loaded from the crate's state file.
     fn set_contracts_state(&mut self, custom_state: Option<Value>) {
-        let state;
-
-        let state_file = Self::deployed_state_file_path();
-        if let Some(custom_state) = custom_state {
-            state = custom_state;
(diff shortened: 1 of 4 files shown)


cw-orch-daemon/src/state.rs
```diff
@@ -357,20 +357,13 @@ pub trait DeployedChains<Chain: CwEnv>: cw_orch_core::contract::Deploy<Chain> {
     /// Set the default contract state for a contract, so that users can retrieve it in their application when importing the library
     /// If a state is provided, it is used for all contracts, otherwise, the state is loaded from the crate's state file.
     fn set_contracts_state(&mut self, custom_state: Option<Value>) {
-        let state;
-
-        let state_file = Self::deployed_state_file_path();
-        if let Some(custom_state) = custom_state {
-            state = custom_state;
```
Buckram123, line 364: It was not patched for custom

cw-orch-daemon/src/state.rs
```diff
@@ -381,8 +374,6 @@ pub trait DeployedChains<Chain: CwEnv>: cw_orch_core::contract::Deploy<Chain> {
             // We try to get the code_id for the contract
             if contract.code_id().is_err() {
                 let code_id = state
-                    .get(env_info.chain_name.clone())
-                    .unwrap_or(&Value::Null)
```
Buckram123, line 385: Those won't be here after patch

Kayanski:
Nice catch, do you need a patch to the released version ?

Buckram123:
> Nice catch, do you need a patch to the released version ?

No, turned out I was debugging wrong thing, haha. Thank you!

PR #509, opened 2024-10-22
Fix/add load state from state file

Kayanski:
This PR aims at adding a method on ChainState to indicate whether the environment load state from state file or is just an in-memory environment.

This is used to load state from file ONLY in the case we need it (Daemon, CloneTesting) and not when we don't need it (Mock)

### Checklist

- [ ] Changelog updated.
- [ ] Docs updated.


Buckram123 (approved):
Ok, this is great! Is there any other cosmwasm envs where we want to set it to `true`?

Kayanski:
> Ok, this is great! Is there any other cosmwasm envs where we want to set it to `true`?

We have: 
- Daemon (true)
- Clone Testing (true)
- Osmosis Test Tube (false)
- Mock (false)

We don't have other envs ?

PR #524, opened 2024-11-12
Added ics hook

Kayanski:
This Pr aims at adding the ics hook ack into cw-orch-interchain ack parsing.
This change is breaking, but we might push the change as non-breaking because the Ack enum is barely used outside of the crate

### Checklist

- [ ] Changelog updated.
- [ ] Docs updated.

### DIFF
--- a/packages/interchain/interchain-core/src/packet.rs
+++ b/packages/interchain/interchain-core/src/packet.rs
@@ -170,6 +170,7 @@ impl<Chain: CwEnv> IndexResponse for NestedPacketsFlow<Chain> {
 }
 
 pub mod success {
+    use crate::ack_parser::IbcHooksAck;
     use crate::{ack_parser::polytone_callback::Callback, tx::TxId};
     use cosmwasm_std::{Binary, Empty, StdError};
     use cw_orch_core::environment::CwEnv;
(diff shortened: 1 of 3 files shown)


packages/interchain/interchain-core/src/packet.rs
```diff
@@ -170,20 +170,25 @@ impl<Chain: CwEnv> IndexResponse for NestedPacketsFlow<Chain> {
...
 pub mod success {
+    use crate::ack_parser::IbcHooksAck;
     use crate::{ack_parser::polytone_callback::Callback, tx::TxId};
     use cosmwasm_std::{Binary, Empty, StdError};
     use cw_orch_core::environment::CwEnv;
     use cw_orch_core::environment::IndexResponse;
 
     /// Contains the result (ack success) associated with various Ibc applications
     #[derive(Debug, PartialEq, Clone)]
+    #[non_exhaustive]
```
Buckram123, line 181: Adding extra field and `non_exhaustive` should be breaking I suppose

Buckram123 (approved):
Perfect, don't forget to update the version

packages/interchain/interchain-core/src/packet.rs
Kayanski: Yes exactly, but as mentionned in the PR description, I'm not sure we should care here, thoughts ?

Buckram123: > Yes exactly, but as mentionned in the PR description, I'm not sure we should care here, thoughts ?

Yeah, sounds good, as long as this enum not used by any other cw-orch package

PR #526, opened 2024-11-20
Added bank send function

Kayanski:
This PR aims at adding the bank send function on cwEnv directly using the TxHandler trait. 
This function is mandatory for blockchain operations and shouldn't need to belong to another trait.

Breaking on cw-orch-daemon

Updates to: 
- Core
- Clone testing
- Mock
- Test tubes

### Checklist

- [ ] Changelog updated.
- [ ] Docs updated.


Kayanski:
> Nice! Are we sure about removing instead of deprecating BankModule trait on cw-orch-interchain?

This was only a test trait, so yeah sure about that

